PR #14, opened 2022-12-15
add DG node as target when DG relation selected

mdroidian:
https://user-images.githubusercontent.com/3792666/207984327-2a0d51a0-80a4-48a7-9de5-ab9a2bf7db34.mp4

### DIFF
--- a/src/utils/registerDiscourseDatalogTranslators.ts
+++ b/src/utils/registerDiscourseDatalogTranslators.ts
@@ -394,12 +394,14 @@ const registerDiscourseDatalogTranslators = () => {
                   dr.source === source ||
                   nodeLabelByType[dr.source] === source)
             );
+          pageNames.push(...Object.values(nodeLabelByType));
           return pageNames.filter((p) =>
             sourcedRelations.some((sr) =>
               matchDiscourseNode({
                 ...nodeByType[sr.target],
                 title: p,
               })
+              || nodeByType[sr.target].text === p
             )
           );
         },
--- a/src/utils/registerDiscourseDatalogTranslators.ts
+++ b/src/utils/registerDiscourseDatalogTranslators.ts
@@ -377,7 +377,7 @@ const registerDiscourseDatalogTranslators = () => {
           ];
         },
         targetOptions: (source) => {
-          let pageNames = getAllPageNames();
+          const pageNames = getAllPageNames();
           const sourcedRelations = discourseRelations
             .flatMap((dr) => [
               { source: dr.source, relation: dr.label, target: dr.destination },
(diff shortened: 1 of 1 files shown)


mdroidian:
@dvargas92495

mdroidian:
But this does seem to highlight an existing bug:

using `query block` gives the wrong datalog query when selecting DG node as target as it forces the user to query
```
Find node Where
node is a Claim
Claim Supported By Evidence
```
as opposed to `query drawer` where a user can target a DG node directly to produce 
```
Find Claim Where
Claim Supported By Evidence
```


https://user-images.githubusercontent.com/3792666/207984825-75fdcc46-9e9a-4ade-b630-1ef230b9f449.mp4

src/utils/registerDiscourseDatalogTranslators.ts
```diff
@@ -394,12 +394,14 @@ const registerDiscourseDatalogTranslators = () => {
                   dr.source === source ||
                   nodeLabelByType[dr.source] === source)
             );
+          pageNames.push(...Object.values(nodeLabelByType));
```
dvargas92495, line 397: If there is an existing page name in the graph called `Claim`, It will come up twice as an option.

we could avoid this by filtering the values before concatting to the array:

```typescript
...Object.values(nodeLabelByType).filter(p => sourcedRelations.some((sr) =>
     nodeByType[sr.target].text === p
));
```

this also removes a check for all the pages that it is not relevant for

dvargas92495: granted, this problem still exists if the `Claim` page itself happens to be a valid discourse node, but that doesn't seem likely.

To protect against that, we could do a `Array.from(new Set(array))` where `array` is our filtered page name list. but this isn't necessary and could be more expensive than its worth

dvargas92495 (commented):
Nice! Left one thought

We can also note the bug you noticed - can try to tackle next week!

mdroidian:
Something like this? https://github.com/dvargas92495/roamjs-query-builder/pull/14/commits/b69051981cdc916e72a87c8abf1b505c3606b500
@dvargas92495

src/utils/registerDiscourseDatalogTranslators.ts
```diff
@@ -394,14 +394,16 @@ const registerDiscourseDatalogTranslators = () => {
...
+          pageNames = pageNames.filter((p) =>
             sourcedRelations.some((sr) =>
               matchDiscourseNode({
                 ...nodeByType[sr.target],
                 title: p,
               })
             )
           );
+          pageNames.push(...Object.values(nodeLabelByType).filter((p) => sourcedRelations.some((sr) => nodeByType[sr.target].text === p)));
+          return pageNames
```
dvargas92495, line 406: you can use `concat` instead of `let`:

```typescript
return pageNames.filter((p) =>
    sourcedRelations.some((sr) =>
        matchDiscourseNode({
                ...nodeByType[sr.target],
                title: p,
        })
    )
).concat(
    ...Object.values(nodeLabelByType).filter(
        (p) => sourcedRelations.some(
            (sr) => nodeByType[sr.target].text === p
        )
    )
)
```

dvargas92495:
also feel free to hit this rerequest review button instead of needing another comment - I receive it as the same notification on my end
Screen Shot 2022-12-15 at 10 07 46 PM

dvargas92495 (approved):
🔥

PR #15, opened 2022-12-15
add query used to when no results found

mdroidian:
@dvargas92495 

https://user-images.githubusercontent.com/3792666/207988330-9efac80b-641d-4d27-b5da-8f151c084122.mp4

### DIFF
--- a/src/components/ResultsView.tsx
+++ b/src/components/ResultsView.tsx
@@ -910,9 +910,37 @@ const ResultsView: typeof window.roamjs.extension.queryBuilder.ResultsView = ({
             <div>Layout `{layout}` is not supported</div>
           )
         ) : (
-          <p className="px-2 py-3 flex justify-between items-center mb-0">
-            <i>No Results Found</i>
-          </p>
+          <>
+            <p className="px-2 py-3 flex justify-between items-center mb-0">
+              <i>No Results Found</i>
+            </p>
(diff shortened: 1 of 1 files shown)


src/components/ResultsView.tsx
```diff
@@ -910,9 +910,37 @@ const ResultsView: typeof window.roamjs.extension.queryBuilder.ResultsView = ({
...
+                    <Button
+                      icon={showContent ? "caret-down" : "caret-right"}
+                      minimal
+                      small
+                      onClick={() => setShowContent(!showContent)}
+                      style={{
+                        marginRight: 4,
+                      }}
+                    />
+                    Showing {paginatedResults.length} of {results.length}{" "}
```
dvargas92495, line 936: Note that because this is the case where there's no results found, this will always be `Showing 0 of 0`, which is redundant with the message above. We could probably just have the carot or combine it with the message above.

No need to do so as part of this task, can handle in a followup if you like. merging!

dvargas92495 (approved):
Nice!!

src/components/ResultsView.tsx
mdroidian: Ah, good point!

Although, as a user, there is something comforting about uniform or familiar interface components.  
Interesting tradeoffs to consider.

PR #18, opened 2022-12-23
add select regex from condition target

mdroidian:
Having some trouble with which characters get escaped on user input for a regex expression when creating `new RegExp` and how to deal with that.

All of these queries work in VSCode on the same input.  I feel that would be a poor user experience.

Let me know your thought @dvargas92495 

image

### DIFF
--- a/src/utils/fireQuery.ts
+++ b/src/utils/fireQuery.ts
@@ -200,6 +200,7 @@ const optimizeQuery = (
   return orderedClauses;
 };
 
+const REGEX_TEST = /{([^}]*)}/;
 const CREATE_DATE_TEST = /^\s*created?\s*(date|time)\s*$/i;
 const EDIT_DATE_TEST = /^\s*edit(?:ed)?\s*(date|time)\s*$/i;
 const CREATE_BY_TEST = /^\s*(author|create(d)?\s*by)\s*$/i;
@@ -318,6 +319,8 @@ const predefinedSelections: PredefinedSelection[] = [
         ? `[:create/time]`
         : EDIT_DATE_TEST.test(field)
         ? `[:edit/time]`
+        : REGEX_TEST.test(field)
+        ? `[:node/title :block/string]`
         : field
         ? `[:block/uid]`
         : `[:node/title :block/uid :block/string]`;
@@ -343,6 +346,8 @@ const predefinedSelections: PredefinedSelection[] = [
         ? getUserDisplayNameById(r?.[":create/user"]?.[":db/id"])
         : field === ":edit/user"
         ? getUserDisplayNameById(r?.[":edit/user"]?.[":db/id"])
+        : REGEX_TEST.test(match)
+        ? new RegExp((match.slice(1,-1))).exec(r?.[":block/string"])?.[0]
         : match
         ? getBlockAttribute(match, r)
         : {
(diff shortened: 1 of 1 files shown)


dvargas92495:
> Having some trouble with which characters get escaped

We can postprocess the user regex, by replacing all instances of `/` with two `//` before passing into `RegExp` constructor.

> I feel that would be a poor user experience.

What would?

mdroidian:
Sorry, that console log wasn't an accurate representation of the regex that was being used. 
Here is what `console.log(new RegExp((match.slice(1, -1))))` prints

For whatever reason, the `\` in the `Tag` selection doesn't get changed

image

So if we replace all `\`, I think that breaks the `Tag` as well.

`console.log((match.slice(1, -1).replace(/\\/g, '\\\\')));`
image

Actually, I think the problem is the `:`
It's not even getting to the `RegExp`.  The time selection `match` isn't logging at all.

I've been trying to edit `NODE_TEST` which fixes this issues but breaks other node selections
`const NODE_TEST = /^node:(\s*[^:]+\s*)(?::({[^}]*})|[^:]+)?$/i;`

mdroidian:
> > I feel that would be a poor user experience.
> 
> What would?

This was in reference to shipping it with some queries working and others not 😀

dvargas92495:
Oh you're right it's the colon messing with the `NODE_TEST` regex. here's something we could try:
```javascript
//                                     node:(any var):(text with no colon OR a regex)
const NODE_TEST = /^node:(\s*[^:]+\s*)(?::((?:[^:]+)|(?:\/.+\/)))?$/i
```

Basically make the last part say, "a colon-less string, or a regex string".

We _might_ also be able to just do this:

```javascript
//                                     node:(any var):(text with no colon OR a regex)
const NODE_TEST = /^node:(\s*[^:]+\s*)(?::.*)?$/i
```

Basically just grab whatever is at the end. this prevents us from having a _fourth_ argument to this convoluted syntax in the future, but that also might be a beneficial constraint from a UX perspective. I'll let you make the call here!

mdroidian:
Unfortunately I couldn't get either of those to work.

`const NODE_TEST = /^node:(\s*[^:]+\s*)(?::((?:[^:]+)|(?:\/.+\/)))?$/i`
Doesn't capture `node:node:{(\d\d?:\d\d)}`
What is the definition of a `regex string`? 
I couldn't seem to figure that out.  It seemed that nearly anything could be a regex string.

`const NODE_TEST = /^node:(\s*[^:]+\s*)(?::.*)?$/i`
Looks like this only returns the first group because of the `?:`

What about something like
`const NODE_TEST = /^node:(\s*[^:]+\s*):(.*)?$/i;`

image

dvargas92495:
> Looks like this only returns the first group because of the ?:

ah then let's just remove the `?:` part!

> What about something like

This won't work, bc it requires that the selection have all three parts. `node:data` however needs to be valid

mdroidian:
image

dvargas92495 (commented):
agh my bad, these two comments were never submitted when you first shared the PR, my last bit of feedback

Also, no need to do it today, enjoy the holidays!

mdroidian:
image
image

dvargas92495 (approved):
nice work!

PR #19, opened 2022-12-29
alias_test, change defaultSelection label if exist

mdroidian:
@dvargas92495 take a look

image

### DIFF
--- a/src/utils/fireQuery.ts
+++ b/src/utils/fireQuery.ts
@@ -200,6 +200,7 @@ const optimizeQuery = (
   return orderedClauses;
 };
 
+const ALIAS_TEST = /^node$/i;
 const REGEX_TEST = /\/([^}]*)\//;
 const CREATE_DATE_TEST = /^\s*created?\s*(date|time)\s*$/i;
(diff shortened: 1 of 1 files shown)


src/utils/fireQuery.ts
```diff
@@ -474,7 +480,7 @@ export const getDatalogQueryComponents = ({
         };
       },
       pull: `(pull ?${returnNode} [:block/string :node/title :block/uid])`,
-      label: "text",
+      label: selections.find(s => s.text === "node")?.label || "text",
```
dvargas92495, line 483: It's a little odd for default values to be based on the input, but I think this is probably the cleanest way to handle this! I think we should send it 🔥

dvargas92495 (approved):

src/utils/fireQuery.ts
dvargas92495: An improvement on this, to make extensible with a configurable `returnNode` (like in Query Drawer), is to do `find(s => s.text === returnNode)`

The second part on that is for somehow `ALIAS_TEST` to take the returnNode into account - this could all be revisited when deciding whether query blocks should have a configurable return node or not

PR #20, opened 2023-01-03
add page embed check and css fix

mdroidian:
@dvargas92495 

before
image

after
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "query-builder",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "description": "Introduces new user interfaces for building queries in Roam",
   "main": "./build/main.js",
   "author": {
--- a/src/components/ResultsView.tsx
+++ b/src/components/ResultsView.tsx
@@ -122,7 +122,9 @@ const ResultHeader = ({
   );
 };
 
+
 const CellEmbed = ({ uid }: { uid: string }) => {
+  const isPage = !!getPageTitleByPageUid(uid) ? "page-embed" : "block-embed"
   const contentRef = useRef(null);
   useEffect(() => {
     window.roamAlphaAPI.ui.components.renderBlock({
@@ -130,7 +132,7 @@ const CellEmbed = ({ uid }: { uid: string }) => {
       el: contentRef.current,
     });
   }, [contentRef]);
-  return <div ref={contentRef} className="roamjs-query-embed" />;
+  return <div ref={contentRef} className={`roamjs-query-embed ${isPage}`} />
 };
 
 const ResultView = ({
--- a/src/components/ResultsView.tsx
+++ b/src/components/ResultsView.tsx
@@ -122,7 +122,6 @@ const ResultHeader = ({
   );
 };
 
-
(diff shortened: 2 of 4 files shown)


src/components/ResultsView.tsx
```diff
@@ -122,15 +122,17 @@ const ResultHeader = ({
   );
 };
 
+
```
dvargas92495, line 125: let's 🔥 this whitespace

src/components/ResultsView.tsx
```diff
@@ -122,15 +122,17 @@ const ResultHeader = ({
...
+  const isPage = !!getPageTitleByPageUid(uid) ? "page-embed" : "block-embed"
   const contentRef = useRef(null);
   useEffect(() => {
     window.roamAlphaAPI.ui.components.renderBlock({
       uid,
       el: contentRef.current,
     });
   }, [contentRef]);
-  return <div ref={contentRef} className="roamjs-query-embed" />;
+  return <div ref={contentRef} className={`roamjs-query-embed ${isPage}`} />
```
dvargas92495, line 135: nice - let's also display the title of the page somewhere in this markup. a simple `<h1 className="rm-page-title">` within the `div` when `isPage` is true should be sufficient

dvargas92495 (commented):
nice! left one bit of feedback

Also - after committing the next bit of changes, can you run `npm version patch` before pushing to this PR? this should set RoamJS Query Builder to a new version number, which will automatically deploy the change to Roam Depot review when merged

src/components/ResultsView.tsx
mdroidian: I'm having trouble getting the `h1` to stay.  It seems `renderBlock` keeps nuking it.
I've tried adding it inside the `useEffect` adding a separate `useEffect` as well as in the `return`

Would this require a larger refactor or am I missing something simple?

mdroidian: or we can nest the `ref={contentRef}` ... one sec ...

src/components/ResultsView.tsx
```diff
@@ -123,14 +123,15 @@ const ResultHeader = ({
 };
 
 const CellEmbed = ({ uid }: { uid: string }) => {
+  const isPage = !!getPageTitleByPageUid(uid) ? "page-embed" : "block-embed"
```
dvargas92495, line 126: Let's make sure we only call `getPageTitleByPageUid(uid)` once - we want to do Roam data lookups as infrequently as possible:
```typescript
const title = getPageTitleByPageUid(uid);
const isPage = !!title ? "page-embed" : "block-embed";
...
<h1 className="rm-title-display">{title}</h1>
```

src/components/ResultsView.tsx
dvargas92495: yup this is how we want to do this - it's fine if the title is not editable.

Let's also add a margin-left to the title and remove the margin-top/bottom

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "query-builder",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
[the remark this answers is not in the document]
mdroidian: How would you like to proceed?

Issue #22, opened 2023-01-08
Sortable Linked References icon disappears


ViTuDia:
Sortable Linked References icon only appears when having the button switched off and on again. (It happens that the icon disappears after reloading Roam while this option having switched on in settings)

<img width="663" alt="Screenshot 2023-01-08 at 15 23 19" src="https://user-images.githubusercontent.com/59163092/211201484-a501ad4c-0b97-40ce-b0cd-8be8f8a00a7a.png">

dvargas92495:
Am able to reproduce, investigating

PR #23, opened 2023-01-10
add toggleSortReferences on load if true

mdroidian:
related to https://github.com/dvargas92495/roamjs-query-builder/issues/22

@dvargas92495, could it be the case that `runSortReferences()` isn't being called on load?

`if (!!extensionAPI.settings.get("sort-references")) {toggleSortReferences(true);}`
seemed to fix it when testing with `Reload Developer Extensions`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -330,7 +330,9 @@ export default runExtension({
       onloadArgs
     );
     const toggleSortReferences = runSortReferences();
-    if (!!extensionAPI.settings.get("sort-references")) {toggleSortReferences(true);}
+    if (!!extensionAPI.settings.get("sort-references")) {
+      toggleSortReferences(true);
+    }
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -330,7 +330,8 @@ export default runExtension({
       onloadArgs
     );
     const toggleSortReferences = runSortReferences();
-
+    if (!!extensionAPI.settings.get("sort-references")) {toggleSortReferences(true);}
```
dvargas92495, line 333: That's probably it!

As an fyi, I'd add the [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) extension to your VSCode workspace. It will auto format the code to look "prettier" on save, based on what we have configured

PR #27, opened 2023-01-27
getEnglishQuery change output number toString

mdroidian:
**Issue this is addressing**
Unable to filter [Discourse Graph Attributes](https://oasis-lab.gitbook.io/roamresearch-discourse-graph-extension/guides/exploring-your-discourse-graph/discourse-attributes)

Although this does fix the Discourse Graph Attribute issue, I haven't put a lot of thought into other possible negative downstream effects.  I tested a regular Attribute that had numbers and a regex selection of numbers, both worked.

Before
image

After
image

### DIFF
--- a/src/utils/fireQuery.ts
+++ b/src/utils/fireQuery.ts
@@ -553,6 +553,8 @@ const getEnglishQuery = (args: FireQueryArgs) => {
                   Object.entries(output).forEach(([k, v]) => {
                     p[label + k] = v;
                   });
+                } else if (typeof output === "number") {
+                  p[label] = output.toString();
                 } else {
                   p[label] = output;
                 }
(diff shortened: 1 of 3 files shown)


src/utils/fireQuery.ts
```diff
@@ -553,6 +553,8 @@ const getEnglishQuery = (args: FireQueryArgs) => {
                   Object.entries(output).forEach(([k, v]) => {
                     p[label + k] = v;
                   });
+                } else if (typeof output === "number") {
+                  p[label] = output.toString();
```
dvargas92495, line 557: I'm concerned that this may break sorting - if you have three results: 1, 11, 2, how does that sort with this change?

mdroidian: Ah good thought.
Apparently it does sort.

No sort
image

Descending
image

Ascending
image

dvargas92495: Sounds good!

PR #29, opened 2023-02-03
Sort ref bug

mdroidian:
@dvargas92495 for [Sort References Bug](https://roamresearch.slack.com/archives/C016N2B66JU/p1675448161078559)

`getPageTitleValueByHtmlElement(sortContainer)` is still required for Daily Notes Page
`getCurrentPageUid()` returns today's date on DNP which is what `/^([0-9]{2}-){2}[0-9]{4}$/.test(currentPageUID)` is looking for.
I feel like this isn't ideal, and it gives a false positive for any day page, but that code still works regardless on those pages.

also `npm version patch` errored, but still went to 1.1.8

image

### DIFF
--- a/src/utils/runSortReferences.ts
+++ b/src/utils/runSortReferences.ts
@@ -184,8 +184,12 @@ const runSortReferences = () => {
       ) => {
         const sidebarWindow = sortContainer.closest(".rm-sidebar-window");
         const currentPageUID = await window.roamAlphaAPI.ui.mainWindow.getOpenPageOrBlockUid();
+        const allSidebarWindows = document.body.querySelectorAll(".rm-sidebar-window");
+        const sidebarWindowIndex = Array.from(allSidebarWindows).indexOf(sidebarWindow);
+        const sidebarWindowData = window.roamAlphaAPI.ui.rightSidebar.getWindows()[sidebarWindowIndex]
         const pageUID = !!sidebarWindow
-          ? getPageUidByPageTitle(sidebarWindow.querySelector("div > a").textContent)
+        //  @ts-ignore Roam has an API inconsistency - https://roamresearch.slack.com/archives/C02TMKXNVS6/p1676389252784269
+          ? sidebarWindowData["page-uid"] || sidebarWindowData["mentions-uid"]
           : !currentPageUID
           ? getPageUidByPageTitle(getPageTitleValueByHtmlElement(sortContainer))
           : currentPageUID;
--- a/src/utils/runSortReferences.ts
+++ b/src/utils/runSortReferences.ts
@@ -2,7 +2,6 @@ import getFullTreeByParentUid from "roamjs-components/queries/getFullTreeByParen
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
 import getPageTitleValueByHtmlElement from "roamjs-components/dom/getPageTitleValueByHtmlElement";
 import getPageTitlesReferencingBlockUid from "roamjs-components/queries/getPageTitlesReferencingBlockUid";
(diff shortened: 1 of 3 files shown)


src/utils/runSortReferences.ts
```diff
@@ -182,11 +183,15 @@ const runSortReferences = () => {
           b: { title: string; time: number }
         ) => number
       ) => {
-        const pageTitle = getPageTitleValueByHtmlElement(sortContainer);
-        if (!pageTitle) {
+        const currentPageUID = getCurrentPageUid();
```
dvargas92495, line 186: The reason we can't use `getCurrentPageUid` is because this won't work for the sort icon in the sidebar. we'll need to derive the block uid from the sort container, returning the current page uid for the main one, or getting it from the rightSidebar methods if it's from the sidebar one.

Also, in general, we should use `window.roamAlphaAPI.ui.mainWindow.getOpenPageOrBlockUid()`, which is a promise the returns the current uid, instead of `getCurrentPageUid`

mdroidian: > sort icon in the sidebar

Can you send me a screenshot of what you are referring to here?

mdroidian: Nevermind, found it.

mdroidian: Both `getOpenPageOrBlockUid()` and `getCurrentPageUid()` seem to work for the sidebar.


https://user-images.githubusercontent.com/3792666/216740126-bc080910-0512-47ca-9754-852061d71c11.mp4

dvargas92495: Right bc you are testing the same page in the sidebar as the one in the main window - test it with a _different_ page in the side bar

mdroidian:
Not super happy with the "Check if sidebar" and "Find pageUID if sidebar" methods, but I didn't see anything helpful for this with `roamAlphAPI.ui.rightSidebar`, nor in `roamjs-components`

(other than maybe modifying `getPageTitleByHtmlElement` to include this functionality)

src/utils/runSortReferences.ts
```diff
@@ -175,18 +175,24 @@ const runSortReferences = () => {
...
           a: { title: string; time: number },
           b: { title: string; time: number }
         ) => number
       ) => {
-        const pageTitle = getPageTitleValueByHtmlElement(sortContainer);
-        if (!pageTitle) {
+        const sidebarWindow = sortContainer.closest(".rm-sidebar-window");
+        const currentPageUID = await window.roamAlphaAPI.ui.mainWindow.getOpenPageOrBlockUid();
+        const pageUID = !!sidebarWindow
+          ? getPageUidByPageTitle(sidebarWindow.querySelector("div > a").textContent)
```
dvargas92495, line 188: I think the "Check if sidebar" is sufficient. Not much else we could do here besides DOM crawling.

"Find pageUID if sidebar" could be more deterministic. we could query select all of the `.rm-sidebar-window` and see what index in that array our `sidebarWindow` is in. Then, we could use `window.roamAlphaAPI.ui.rightSidebar.getWindows()` to get the pageUid.

```typescript
const sidebarWindow = sortContainer.closest(".rm-sidebar-window");
const allWindows = document.body.querySelectorAll(".rm-sidebar-window");
const sidebarWindowIndex = Array.from(allWindows).indexOf(sidebarWindow);
const sidebarWindowData = window.roamAlphaAPI.ui.rightSidebar.getWindows()[sidebarWindowIndex]
// @ts-ignore we need to ignore typescript here for now bc Roam has an API inconsistency - see my latest msg in #developers channel
const pageUID = sidebarWindowData["page-uid"] || sidebarWindowData["mentions-uid"]; 
```

dvargas92495 (approved):

PR #44, opened 2023-02-16
Support adding aliases to Query Blocks closes #30

### LINKED ISSUE #30: Support adding aliases to Query Blocks
Then the QUERYBUILDER smartblock command should be able to reference these aliases from the block or from the page name

mdroidian:


mdroidian:
doesn't deal with use case of multiple same named queries (just takes the first one)

- some block
  - `{{query block:myQuery}}`
- some other block 
  - `{{query block:myQuery}}`

Page: `queries/myQuery`

src/index.ts
```diff
@@ -422,8 +422,20 @@ export default runExtension({
...
+                 [and [?b :block/string ?s] [[clojure.string/includes? ?s "{{query block:${queryRef}}}"]] ]
+                 ${getQueryPages(extensionAPI).map(
+                   (p) =>
+                     `[and [?b :node/title ?t] [[re-pattern "^${p.replace(
+                       /\*/,
+                       queryRef
+                     )}$"] ?regex] [[re-find ?regex ?t]]]`
+                 )}
+            ]]`
+          )[0]?.toString();
```
dvargas92495, line 437: I was hoping for the alias to be editable in the UI but that may not be necessary. At the very least, it should be viewable in the UI somehow. If it's editable, we would have to build out:
```
- {{query block}}
    - alias
        - example
```

which might be more trouble than it's worth. Your solution handles the issue rather quickly. Thoughts?

In terms of retrieval, this is 🎯

src/index.ts
```diff
@@ -422,8 +422,20 @@ export default runExtension({
...
+          const queryRef = variables[arg] || arg;
+          const aliasOrPage = window.roamAlphaAPI.data.fast.q(
+            `[:find ?uid :where [?b :block/uid ?uid] [or-join [?b] 
+                 [and [?b :block/string ?s] [[clojure.string/includes? ?s "{{query block:${queryRef}}}"]] ]
+                 ${getQueryPages(extensionAPI).map(
+                   (p) =>
+                     `[and [?b :node/title ?t] [[re-pattern "^${p.replace(
+                       /\*/,
+                       queryRef
+                     )}$"] ?regex] [[re-find ?regex ?t]]]`
```
dvargas92495, line 434: why not a straight comparison `[and [?b :node/title "^${p.replace(/\*/, queryRef)}$"]]`?

I also think we'll want to add a branch for comparing the title directly: `[and [?b :node/title "${queryRef}"]]`

src/index.ts
dvargas92495: Reflecting on it more, I think I prefer the approach you took over the one I had in mind. Only addition I'd make is to make that alias viewable in the query block somehow

src/index.ts
mdroidian: What is the use case for comparing the title directly?

src/index.ts
mdroidian: I could see this being nice to unify where to find alias across `Query Blocks`, `Query Drawer`, and `Query Pages`!

mdroidian:
Not super happy with the default `Query Alias`  `<span>`
It doesn't seem intuitive as to what it's function is.

I'll come back to this.

src/components/QueryEditor.tsx
```diff
@@ -494,6 +500,38 @@ const QueryEditor: typeof window.roamjs.extension.queryBuilder.QueryEditor = ({
...
+                setIsEditingLabel(false);
+              }
+            }}
+            autoFocus
+            rightElement={
+              <Button
+                minimal
+                icon={"confirm"}
+                onClick={() => {
+                  updateBlock({ uid: parentUid, text: `{{query block:${label}}}` });
```
dvargas92495, line 521: what does this UX look like? does the component flicker or is it smooth?

src/index.ts
dvargas92495: Agreed! Very nicely done.

src/components/QueryEditor.tsx
```diff
@@ -494,6 +500,39 @@ const QueryEditor: typeof window.roamjs.extension.queryBuilder.QueryEditor = ({
...
+            }
+          />
+        ) : (
+          <span 
+            style={{display: "inline-block"}}
+            tabIndex={-1} 
+            onClick={() => setIsEditingLabel(true)}
+          >
+            {!!label ? label : "Query Alias"}
+          </span>
```
dvargas92495, line 534: > Not super happy with the default Query Alias <span>

When there's no alias, what about a span the is italicized, low opacity, lower font, that says `Click to edit alias` ?

src/components/QueryEditor.tsx
mdroidian: Seems smooth.

https://user-images.githubusercontent.com/3792666/219730603-51b4925c-158c-4eab-96a1-4d339b28d82f.mp4

src/components/QueryEditor.tsx
mdroidian: Yeah, not bad!

image

I was thinking that if majority of users / queries won't use this feature then to move it to the menu.
But then the user would have to click once (or twice) to see the alias.

But I think the muted text is probably acceptable.

src/index.ts
mdroidian: Let me know if you want that done in this PR or if we should ship what we have and look at unifying it later.

I've converted the PR from a draft either way.

dvargas92495: no need for it to be done now this looks good to ship

dvargas92495 (approved):
:fire:

PR #48, opened 2023-02-20
add showAlias props closes #47

### LINKED ISSUE #47: Alias on Query Drawer / Query Pages
While making the docs for #44, I noticed that missed looking at `Query Drawer` / `Query Pages`

`Query Drawer` doesn't serve a purpose
image

I thought the same would be true for `Query Pages` ... but ...
image

Where is `{{query block:whyDoesThisWork}}` stored for `Query Pages`?

# Possible solution
adding this check to QueryEditor.tsx to remove it from the `Query Drawer`

`const isQueryDrawer = getPageTitleByPageUid(parentUid) === "roam/js/query-builder/drawer";`

mdroidian:
`true` for `Query Blocks` and `Query Pages`

### DIFF
--- a/src/components/QueryPage.tsx
+++ b/src/components/QueryPage.tsx
@@ -176,6 +176,7 @@ const QueryPage = ({
         {isEdit && (
           <>
             <QueryEditor
+              showAlias={true}
               parentUid={pageUid}
               onQuery={() => setIsEdit(false)}
               defaultReturnNode={defaultReturnNode}
--- a/src/components/QueryPage.tsx
+++ b/src/components/QueryPage.tsx
@@ -3,8 +3,8 @@ import React, { useCallback, useEffect, useMemo, useRef, useState } from "react"
 import fireQuery from "../utils/fireQuery";
 import parseQuery from "../utils/parseQuery";
 import type {
-  QueryPageComponent,
-  Result as SearchResult,
+  ExportTypes,
+  Result,
 } from "roamjs-components/types/query-builder";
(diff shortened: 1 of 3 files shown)


src/components/QueryPage.tsx
```diff
@@ -176,6 +176,7 @@ const QueryPage = ({
         {isEdit && (
           <>
             <QueryEditor
+              showAlias={true}
```
dvargas92495, line 179: I thought we wanted this only for query block and not query page?

mdroidian:
It turns out `alias` still works for `Query Pages` (stores the string in db without displaying it, I suppose), so having it there is probably more intuitive for the end user (staying consistent with `Query Block`)

But the biggest reason I left it in is that I only found the `<QueryEditor>` component called twice, in `QueryPage.tsx` and `QueryDrawer.tsx`.  It looks like `Query Block` and `Query Page` share the component in `QueryPage.tsx`.

I could be wrong on that.  So if there is a better way, let me know. 👍

dvargas92495:
Here is the different entrypoints for Query Blocks vs Query Pages: https://github.com/dvargas92495/roamjs-query-builder/blob/e9b9f47fb824086c195dd9b104c011b940580e3f/src/components/QueryPage.tsx#L213-L235

storing `:block/string` in page objects in Roam could cause unintended consequences down the road, like a page being misinterpreted as a block or vice versa.

Let's use the entry point above to only show alias for query blocks for now

dvargas92495 (approved):
:fire:

PR #49, opened 2023-02-21
SamePage Cross Notebook Example

dvargas92495:
DO NOT MERGE

PR is exploratory while we refine the cross notebook query API

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,6 @@ import getUidsFromId from "roamjs-components/dom/getUidsFromId";
 import { renderQueryBuilder } from "./components/QueryBuilder";
 import runExtension from "roamjs-components/util/runExtension";
 import addStyle from "roamjs-components/dom/addStyle";
-import getSubTree from "roamjs-components/util/getSubTree";
 import getPageTitleValueByHtmlElement from "roamjs-components/dom/getPageTitleValueByHtmlElement";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
 import {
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -500,6 +503,29 @@ export default runExtension({
...
+        window.roamAlphaAPI.ui.commandPalette.addCommand({
+          label: "Cross notebook request",
+          callback: () => {
+            listNotebooks().then(({ notebooks }) => {
+              const targets = notebooks
+                .map((n) => n.uuid)
+                .filter((n) => n.startsWith("8a80"));
+              console.log("targets", targets);
+              sendNotebookRequest({
+                request: { hello: "world" },
```
dvargas92495, line 517: We could delete this whole command

src/utils/fireQuery.ts
```diff
@@ -597,6 +602,28 @@ const fireQuery: FireQuery = async (args) => {
       console.log("Query to Roam:");
       console.log(query);
     }
+    const response = await getSamePageAPI().then(
+      ({ sendNotebookRequest, listNotebooks }) => {
+        return listNotebooks().then(({ notebooks }) => {
+          const targets = notebooks
+            .map((n) => n.uuid)
+            .filter((n) => n.startsWith("8a80"));
+          console.log("targets", targets);
```
dvargas92495, line 611: Need to instead allow selection of targets rather than this hard code

src/utils/fireQuery.ts
```diff
@@ -597,6 +602,28 @@ const fireQuery: FireQuery = async (args) => {
...
+          let respondedOnce = false;
```
dvargas92495, line 612: fire query needs to accept a command that allows updating of responses `onResponse` - that way we also don't need to await on the API

PR #50, opened 2023-02-21
References title bug

dvargas92495:
image

### DIFF
--- a/src/utils/conditionToDatalog.ts
+++ b/src/utils/conditionToDatalog.ts
@@ -399,10 +399,10 @@ const translator: Record<string, Translator> = {
         arguments: [
           { type: "variable", value: source },
           { type: "constant", value: ":block/refs" },
-          { type: "variable", value: target },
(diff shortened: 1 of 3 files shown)


src/utils/conditionToDatalog.ts
```diff
@@ -399,10 +399,10 @@ const translator: Record<string, Translator> = {
         arguments: [
           { type: "variable", value: source },
           { type: "constant", value: ":block/refs" },
-          { type: "variable", value: target },
+          { type: "variable", value: `${target}-Ref` },
         ],
       },
-      ...getTitleDatalog({ source: target, target }),
+      ...getTitleDatalog({ source: `${target}-Ref`, target }),
```
dvargas92495, line 405: the problem was that we had two different variables using the same name: `[?log :block/refs ?log]`.

Adding the suffix should differentiate

mdroidian (approved):
Makes sense.  Looks good 👍

PR #55, opened 2023-04-01
Incorporating SamePage into QueryBuilder formally

dvargas92495:


dvargas92495:
Adding @mdroidian for visibility, no review yet.

By querying from SamePage, Query Builder will be able to:
- Offload querying work to the backend, resolving performance issues
- Query from other notebooks the user has access to (Obsidian, LogSeq, Notion, Github, etc etc)

.github/workflows/main.yaml
```diff
@@ -21,4 +21,4 @@ jobs:
       - name: install
         run: npm install
       - name: build
-        run: npx roamjs-scripts build --labs
+        run: npx samepage build
```
dvargas92495, line 24: I'd like to start slowly phasing out `roamjs-scripts` in favor of `samepage`, which will come bundled with `roamjs-components` as part of `@samepage/scripts`

src/components/Charts.tsx
```diff
@@ -2,6 +2,8 @@ import React from "react";
 import { Chart, AxisOptions, AxisOptionsBase } from "react-charts";
 import { Result } from "roamjs-components/types/query-builder";
 
+type ChartData = [Result[string], Result[string]];
```
dvargas92495, line 5: this is an example of what not to do in a PR.

The upgrading of `tsconfig.json` (https://github.com/dvargas92495/roamjs-query-builder/pull/55/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0) that happened as a byproduct of going from `roamjs-scripts` to `@samepage/scripts` meant a bunch of typescript rules that weren't previously enforced are now enforced, which led to a bunch of cleanup work throughout the PR.

I'll highlight any relevant changes that weren't a part of this.

src/components/CytoscapePlayground.tsx
```diff
@@ -49,8 +49,9 @@ import getDiscourseRelationTriples from "../utils/getDiscourseRelationTriples";
 import renderWithUnmount from "roamjs-components/util/renderWithUnmount";
```
dvargas92495, line 49: would skip this file

src/components/SendQueryRequest.tsx
```diff
@@ -1,152 +0,0 @@
-import {
```
dvargas92495, line 1: Ripping out all of the old SamePage stuff to instead use the cross-app request protocol that is in development

src/index.ts
```diff
@@ -569,6 +563,18 @@ export default runExtension({
...
+      api.listNotebooks().then(({ notebooks }) => {
+        registerDatalogTranslator({
+          key: "is in notebook",
+          callback: () => [],
+          isVariable: true,
+          placeholder: `Roam ${window.roamAlphaAPI.graph.name}`,
+          targetOptions: notebooks.map((n) => `${n.appName} ${n.workspace}`),
+        });
+      });
+    });
```
dvargas92495, line 576: with samepage enabled, Query builder will have access to a new `is in notebook` relation

src/utils/fireQuery.ts
```diff
@@ -576,15 +580,20 @@ export const getDatalogQuery = ({
...
+      .then((api) =>
+        api.postToAppBackend<{ results: QueryResult[] }>("query", args)
+      )
+      .then((r) => r.results)
+      .catch((e) => {
+        console.error("Error from SamePage:");
+        console.error(e.message);
+        return [];
+      });
+  }
```
dvargas92495, line 596: this is prob the most important change in the PR

src/index.ts
mdroidian: Could you remove the notebook limit on my samepage account?

mdroidian: If I use `node is in notebook someNotebook`, should that  only return nodes in `someNotebook`?

mdroidian (commented):
Maybe we can walk through this on Monday?
I'm not entirely sure if this is supposed to just be a backend change or some frontend changes as well?
I did notice the `is in notebook`, but the callback is blank, does that mean it doesn't do anything yet?
I also noticed the checkbox with the SamePage logo, but I'm not sure what that does either.  Does that offload querying work to the backend?

mdroidian:
Also, could you remove the notebook limit on my samepage account so I could test in a few more places?  Thanks.

dvargas92495:
I need the roamjs-components bump in the tldraw PR, merging

PR #56, opened 2023-04-07
Remove Slashes from Datalog vars

dvargas92495:


src/utils/compileDatalog.ts
```diff
@@ -4,7 +4,7 @@ import type {
   DatalogClause,
 } from "roamjs-components/types/native";
 
-const toVar = (v = "undefined") => v.replace(/[\s"()[\]{}]/g, "");
+const toVar = (v = "undefined") => v.replace(/[\s"()[\]{}/\\]/g, "");
```
dvargas92495, line 7: can you pull and verify it this works for those queries?

mdroidian (changes requested):
`Unmatched delimiter ].`

dvargas92495:
Can you paste in the query that's resulting in that error? I'm failing to repro

mdroidian:
```
Find node Where
node is a Question
```

```
[:find
  (pull ?node [:block/string :node/title :block/uid])
  (pull ?node [:block/uid])
:where
  [(re-pattern "^\\[\\[QUE\\]\\] - (.*?)$") ?/^\\QUE\\-.*?$/-regex]
  [?node :node/title ?node-Title]
  [(re-find ?/^\\QUE\\-.*?$/-regex ?node-Title)]
]
```

mdroidian:
Sorry, posted wrong query.  Edited previous comment.

dvargas92495:
The problem seems to be in the string, not the getVar function. Namely:

`"^\\[\\[QUE\\]\\] - (.*?)$"`

should be

`"^\\\\[\\\\[QUE\\\\]\\\\] - (.*?)$"`

dvargas92495:
Is it not hitting this replace? https://github.com/dvargas92495/roamjs-query-builder/blob/SLASHES/src/utils/conditionToDatalog.ts#L86

mdroidian:
Sorry, I can't seem to test it right now.  I'm having some weird issue.

Whenever I load query builder extension via `Developer Extensions` I'm was getting multiple of the following error:
`ERROR in load-extension: invalid ext-id:  null`

I loaded up a new graph and get this error:
`app-graph.js:68 Uncaught (in promise) Error: Can not register `navigator` for `core` since `navigator` already exists in the prototype and can not be overridden`

And Chrome doesn't register query builder in sources
image

I've tried your branch, my fork, and an old commit on my fork. 🤷‍♂️

To top it off, our power it out, so my laptop is about to die. 😣

mdroidian:
running `npx roamjs-scripts dev --depot` instead of `npm start` seemed to do it

mdroidian:
> Is it not hitting this replace? https://github.com/dvargas92495/roamjs-query-builder/blob/SLASHES/src/utils/conditionToDatalog.ts#L86

I believe it is hitting it
image

dvargas92495:
Ok I was able to reproduce both issues:
- delimiter was solved by adding the `^` to the exclusion regex
- build issue solved by bumping up roamjs-scripts to latest

mdroidian (approved):

PR #58, opened 2023-04-07
Adding an optional limit arg to the QUERYBUILDER cmd

dvargas92495:
Something I'm looking to add for agency's purpose.

I'm gonna be running a QB table _per_ client. Then, I want my DNP to return just the _first_ item from each board

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -417,7 +417,7 @@ export default runExtension({
     registerSmartBlocksCommand({
       text: "QUERYBUILDER",
       delayArgs: true,
-      help: "Run an existing query block and output the results.\n\n1. The reference to the query block\n2. The format to output each result",
+      help: "Run an existing query block and output the results.\n\n1. The reference to the query block\n2. The format to output each result\n3. (Optional) The number of results returned",
       handler:
         ({ proccessBlockText, variables }) =>
         (arg, ...args) => {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -420,7 +420,12 @@ export default runExtension({
       help: "Run an existing query block and output the results.\n\n1. The reference to the query block\n2. The format to output each result",
       handler:
         ({ proccessBlockText, variables }) =>
-        (arg, format = "(({uid}))") => {
+        (arg, ...args) => {
+          const lastArg = args[args.length - 1];
+          const lastArgIsLimitArg = !Number.isNaN(Number(lastArg));
+          const { format, limit } = lastArgIsLimitArg
+            ? { format: args.slice(0, -1).join(","), limit: Number(lastArg) }
+            : { format: args.join(","), limit: 0 };
```
dvargas92495, line 428: this complicated mess is a by product of ordering mattering in Sbs, which is why the next evolution of SBs in SamePage will have order agnostic parameters.

Basically, we want the format argument to be able to support accepting commas, but we can't put the limit argument before it without breaking existing workflows. So, we only assume it is if the last arg is a number

mdroidian: 👍

mdroidian (approved):
Would it be worth adding a line to:
https://github.com/dvargas92495/roamjs-query-builder/blob/5351ed89f84748f1c68ea683ae9aea3f735bbbeb/src/index.ts#L420

```
Run an existing query block and output the results.\n\n1. The reference to the query block\n2. The format to output each result\n3. (Optional) The number of results returned
```

PR #62, opened 2023-04-11
Supporting Inputs

dvargas92495:
If you want to create a backlog for multiple clients or users, you need to create separate query builder instances for each one.

This allows users to only have one query builder instance that can be used across multiple values for a particular field.

We're starting with the SB use case, but I still need to go through all of the usages of `fireQuery` to see what can apply.

Adding you to the PR now for visibility, no need for review yet

### DIFF
--- a/src/utils/fireQuery.ts
+++ b/src/utils/fireQuery.ts
@@ -13,10 +13,11 @@ import apiPost from "roamjs-components/util/apiPost";
 import type { Condition, Selection } from "./types";
 import getSamePageAPI from "@samepage/external/getSamePageAPI";
 
-type QueryArgs = {
+export type QueryArgs = {
   returnNode: string;
   conditions: Condition[];
   selections: Selection[];
+  inputs?: Record<string, string | number>;
 };
 
 type FireQueryArgs = QueryArgs & {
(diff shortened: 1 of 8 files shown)


src/utils/fireQuery.ts
```diff
@@ -522,27 +527,20 @@ export const getDatalogQueryComponents = ({
...
-const getQueryStringAndFormatter = (args: FireQueryArgs) => {
-  const parts = getDatalogQueryComponents(args);
+  const find = definedSelections.map((p) => p.pull).join("\n  ");
+  const where = whereClauses.map((c) => compileDatalog(c, 0)).join("\n  ");
   return {
-    query: getDatalogQuery(parts),
+    query: `[:find\n  ${find}\n${
+      expectedInputs.length
+        ? `  :in $ ${expectedInputs.map((i) => `?${i}`).join(" ")}\n`
+        : ""
```
dvargas92495, line 536: We're finally taking advantage of datalog's `:in` with these changes

dvargas92495:
@mdroidian ready for review

mdroidian:
Can you give an example of how you would use these changes?

mdroidian (approved):
🔥

PR #64, opened 2023-04-12
Exploring Tldraw as cytoscape replacement

dvargas92495:


.gitignore
```diff
@@ -10,3 +10,5 @@ extension.css
 report.html
 stats.json
 analyze.txt
+*.woff2
+*.woff
```
dvargas92495, line 14: heads up as this will be annoying for a bit. Tldraw generates a bunch of `.woff` files, and because we copy everything in `dist` to `root` bc that's where Roam expects it, it will clutter your top level folder

src/components/TldrawCanvas.tsx
```diff
@@ -0,0 +1,107 @@
...
+            tldrawApps[title] = app;
+          }
+        }}
+        onPersist={(app) => {
+          setInputSetting({
+            blockUid: pageUid,
+            key: "State",
+            value: window.btoa(JSON.stringify(app.document)),
+          });
+        }}
```
dvargas92495, line 78: so far, the only feature from playground we've brought over is state persistance in Roam. We'll need to be diligent about how to port the rest over

mdroidian (commented):
This is awesome!  Can't wait! 🚀

A few things I noticed:

Things like the the dark mode not having a dark background and the selection box being completely black make me wonder if the stylesheet wasn't fully applied.

For the selection box, it looks like `--tl-selectFill` isn't defined (should be transparent) in `.t1-bounds-bg`

Also, having all of the fonts in `root` seems odd to me.
EDIT: Nevermind, I missed that comment.

But none of these hamper being able to use it, so I'm not requesting changes, just noting.

I'll let you decide and merge. 👍

PR #66, opened 2023-04-16
add global text search filter (closes #52)

### LINKED ISSUE #52: Text search filters on individual columns

mdroidian:
WIP
- [ ] fix initial load / toggle recognizing settings.searchFilter
- [ ] only match columns, not uids in `postProcessResults`
image

### DIFF
--- a/src/components/ResultsView.tsx
+++ b/src/components/ResultsView.tsx
@@ -37,6 +37,8 @@ import MenuItemSelect from "roamjs-components/components/MenuItemSelect";
 import { RoamBasicNode } from "roamjs-components/types";
 import { render as renderToast } from "roamjs-components/components/Toast";
 import { ExportTypes, QBClauseData } from "../utils/types";
+import updateBlock from "roamjs-components/writes/updateBlock";
+import getFirstChildUidByBlockUid from "roamjs-components/queries/getFirstChildUidByBlockUid";
 
 type Sorts = { key: string; descending: boolean }[];
 type FilterData = Record<string, Filters>;
@@ -424,6 +426,7 @@ const ResultsView: ResultsViewComponent = ({
   const [pageSize, setPageSize] = useState(settings.pageSize);
   const pageSizeTimeoutRef = useRef(0);
   const [views, setViews] = useState(settings.views);
+  const [searchFilter, setSearchFilter] = useState("");
 
   const preProcessedResults = useMemo(
     () => (resultFilter ? results.filter(resultFilter) : results),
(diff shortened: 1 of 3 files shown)


src/components/ResultsView.tsx
```diff
@@ -485,156 +499,202 @@ const ResultsView: ResultsViewComponent = ({
...
-            isEditRandom ? (
-              <div className="relative w-72 p-4">
-                <h4 className="font-bold flex justify-between items-center">
-                  Get Random
-                  <Button
-                    icon={"small-cross"}
-                    onClick={() => setIsEditRandom(false)}
-                    minimal
-                    small
+      <div className="relative">
```
mdroidian, line 556: most of the changes in this file are from adding this `div` and `prettier`

the only thing added after this line is the menu item at line 789

mdroidian:
Take a look 👀

One thing I've noticed that could be an issue:
There is no indicator if a filter has been applied if it's closed.  

We could do something like this, but I'm not a big fan.  Maybe it'll grow on me.
Let me know if you have thoughts on that.

image

src/utils/postProcessResults.ts
```diff
@@ -68,11 +68,13 @@ const postProcessResults = (
     })
     .filter((r) => {
       return settings.searchFilter
-        ? Object.values(r).some((value) =>
-            String(value)
-              .toLowerCase()
-              .includes(settings.searchFilter.toLowerCase())
-          )
+        ? Object.keys(r)
+            .filter((key) => !key.endsWith("-uid") || key !== "uid")
```
dvargas92495, line 69: I think you want just `!key.endsWith("-uid")`. Not only is the second condition redundant with the first, the key `foo-uid` will actually return `true` because it'll satisfy the second part of this `||`

dvargas92495:
> Let me know if you have thoughts on that.

One thing I was thinking about is how I want the Table Alias viewable on this Results View, and how it should probably be as a top line. We could add whatever the ongoing search is to that top line as well.

An alternative to both would be to have like an asterisk on the three dot menu whenever the results are filtered

src/utils/postProcessResults.ts
mdroidian: ah, right.  I changed it to `&&`

mdroidian:
> > Let me know if you have thoughts on that.
> 
> One thing I was thinking about is how I want the Table Alias viewable on this Results View, and how it should probably be as a top line. We could add whatever the ongoing search is to that top line as well.
> 
> An alternative to both would be to have like an asterisk on the three dot menu whenever the results are filtered

Interesting.  The top line could be something to work towards.  I'm not sure how an asterisk would look on top of three dots, though.  

I didn't see any BP icons that looked good off the top, so I tried the same red as filter has.

image

src/index.ts
```diff
@@ -51,6 +51,10 @@ export default runExtension({
     outline-width: 2px;
 }
 
+.bp-warning, .bp-warning.bp3-button .bp3-icon {
```
mdroidian, line 54: added because style was being overridden on the `button`

dvargas92495 (approved):

PR #67, opened 2023-04-17
Creating and removing discourse nodes to/from canvas

dvargas92495:
https://www.loom.com/share/b249d2d7a89e4d09bce9c349694ed0b1

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -19,6 +19,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: install
-        run: npm install
+        run: npm install -f
       - name: build
         run: npx samepage build
--- a/src/components/TldrawCanvas.tsx
+++ b/src/components/TldrawCanvas.tsx
@@ -4,10 +4,9 @@ import isFlagEnabled from "../utils/isFlagEnabled";
 import {
   Tldraw,
   App as TldrawApp,
-  TldrawUiContextProvider,
-  useLocalSyncClient,
   TLInstance,
   TLUser,
+  TldrawEditorConfig,
 } from "@tldraw/tldraw";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
 import getBasicTreeByParentUid from "roamjs-components/queries/getBasicTreeByParentUid";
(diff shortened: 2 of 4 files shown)


.github/workflows/main.yaml
```diff
@@ -19,6 +19,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: install
-        run: npm install
+        run: npm install -f
```
dvargas92495, line 22: this really sucks

Roam is on React 17, but tldraw is on react 18. so we have to force install dependencies each time for now since they are technically incompatible

src/components/TldrawCanvas.tsx
```diff
@@ -1,12 +1,60 @@
...
+  menuItem,
+  OnDoubleClickHandler,
+  TLTranslationKey,
+  OnTranslateEndHandler,
+} from "@tldraw/tldraw";
+import {
+  Button,
+  Classes,
+  Dialog,
+  // Icon,
```
dvargas92495, line 36: I have a local copy of `CytoscapePlayground` that I'm gradually deleting and migrating over time. When that file is empty, migration will be complete. For things I want to leave for later, they've been marked `// TODO`

mdroidian (commented):
Your Loom video seems to have no audio, FYI.

I'm also getting some weird interactions, Canvas not loading, `{query blocks}` not loading, etc.

https://user-images.githubusercontent.com/3792666/233746704-1a553edd-706e-4fe3-adb8-4dd35a7d1a6a.mp4


https://user-images.githubusercontent.com/3792666/233746466-6247ca41-d25d-422c-92b2-72ceae7d536f.mp4

dvargas92495:
https://www.loom.com/share/505597aa0f924bfaa7aa6af203c51695

Not sure why the previous loom had no volume... thanks for the heads up

mdroidian (approved):
I just tried it on two different graphs and the issue wasn't there today.  Very odd 🤷‍♂️
The `signia` / `navigator` error only seemed to show up on reload (not on first load).
I've noticed the `navigator` error for some time (months), but the `signia` one is new.

dvargas92495:
The signia one is a known one related to TLDraw - Doesn't appear to yet have user facing implications.

The navigator error I think we should let be as it relates to cytoscape, which we are actively replacing with TLDraw

PR #72, opened 2023-04-28
Fix Click Handling and SVG rendering

dvargas92495:


package.json
```diff
@@ -30,7 +30,8 @@
     "react-charts": "^3.0.0-beta.48",
     "react-in-viewport": "^1.0.0-alpha.20",
     "react-vertical-timeline-component": "^3.5.2",
-    "roamjs-components": "^0.77.14"
+    "roamjs-components": "^0.77.14",
+    "signia-react": "^0.1.1"
```
dvargas92495, line 34: this is just a library `tldraw` uses, that we're exposing up

mdroidian (approved):
Nice.

One thing I noticed was a significant amount of "dead space" on my discourse nodes compared to rectangles.  I'd have to click like right in the middle.

And I figured that black text node is just a leftover from the first commit?  It was just a "Text node", no corresponding discourse node.

https://user-images.githubusercontent.com/3792666/235270396-548a117f-6c47-4e18-85de-eb4d7dc5ba37.mp4

dvargas92495:
Text nodes are actually supported in the original cytoscape playground - they're meant to represent ordinary roam pages/blocks.

What wasn't expected was the Error, which I just pushed a fix to address, nice find

PR #73, opened 2023-05-02
Supporting action view types that can be used for inserting into curr…

dvargas92495:
…ent canvas/page

### DIFF
--- a/src/components/QueryDrawer.tsx
+++ b/src/components/QueryDrawer.tsx
@@ -40,46 +40,35 @@ const SavedQuery = ({
   uid,
   isSavedToPage = false,
   onDelete,
-  resultsReferenced,
   clearOnClick,
-  setResultsReferenced,
   editSavedQuery,
   initialResults,
-  onRefresh,
 }: {
   uid: string;
   onDelete?: () => void;
   isSavedToPage?: boolean;
-  resultsReferenced: Set<string>;
   clearOnClick: (s: string) => void;
-  setResultsReferenced: (s: Set<string>) => void;
   editSavedQuery: (s: string) => void;
   initialResults?: Result[];
-  onRefresh: () => void;
 }) => {
   const [results, setResults] = useState<Result[]>(initialResults || []);
-  const resultFilter = useCallback(
-    (r: Result) => !resultsReferenced.has(r.text),
-    [resultsReferenced]
-  );
   const [minimized, setMinimized] = useState(!isSavedToPage && !initialResults);
   const [initialQuery, setInitialQuery] = useState(!!initialResults);
   const [label, setLabel] = useState(() => getTextByBlockUid(uid));
   const [isEditingLabel, setIsEditingLabel] = useState(false);
   const [error, setError] = useState("");
-  useEffect(() => {
-    if (!initialQuery && !minimized) {
-      setInitialQuery(true);
+  const resultsInViewRef = useRef<Result[]>([]);
+  const refresh = useCallback(
+    () =>
       fireQuery(parseQuery(uid))
         .then(setResults)
         .catch(() => {
           setError(
             `Query failed to run. Try running a new query from the editor.`
           );
-        });
-    }
-  }, [initialQuery, minimized, setInitialQuery, setResults, uid]);
-  const resultsInViewRef = useRef<Result[]>([]);
+        }),
+    [uid, setResults, setError]
+  );
   return (
     <div
       style={{
@@ -91,7 +80,7 @@ const SavedQuery = ({
     >
       <ResultsView
         parentUid={uid}
-        onRefresh={onRefresh}
+        onRefresh={refresh}
         header={
           error ? (
             <div className="text-red-700 mb-4">{error}</div>
@@ -124,7 +113,7 @@ const SavedQuery = ({
                   {label}
                 </span>
               )}
-              <div className="mr-14 mt-2">
+              <div className="mr-16 mt-2">
                 <Tooltip content={"Edit"}>
                   <Button
                     icon={"edit"}
@@ -208,12 +197,6 @@ const SavedQuery = ({
                           resultsInViewRef.current.map((r) => {
                             clearOnClick?.(r.text || "");
                           });
-                          setResultsReferenced(
-                            new Set([
-                              ...Array.from(resultsReferenced),
-                              ...resultsInViewRef.current.map((r) => r.text),
-                            ])
-                          );
                         }}
                       />
                     </Tooltip>
@@ -249,7 +232,13 @@ const SavedQuery = ({
                     <Tooltip content={minimized ? "Maximize" : "Minimize"}>
                       <Button
                         icon={minimized ? "maximize" : "minimize"}
-                        onClick={() => setMinimized(!minimized)}
+                        onClick={() => {
+                          if (!initialQuery && minimized) {
+                            setInitialQuery(true);
+                            refresh().finally(() => setMinimized(false));
+                          }
+                          setMinimized(!minimized);
+                        }}
                         active={minimized}
                         minimal
                       />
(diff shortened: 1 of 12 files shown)


dvargas92495:
https://www.loom.com/share/14b842ec259f4819bb189a20204dca09

src/components/QueryDrawer.tsx
```diff
@@ -40,46 +40,35 @@ const SavedQuery = ({
...
-  setResultsReferenced: (s: Set<string>) => void;
   editSavedQuery: (s: string) => void;
   initialResults?: Result[];
-  onRefresh: () => void;
 }) => {
   const [results, setResults] = useState<Result[]>(initialResults || []);
-  const resultFilter = useCallback(
-    (r: Result) => !resultsReferenced.has(r.text),
-    [resultsReferenced]
-  );
```
dvargas92495, line 64: A big part of these changes was trying to get rid of custom logic that query editor/results view had to handled.

Here, `resultsReferenced` could now be handled with the `is in canvas` and `is referenced by block in page with title` conditions

src/components/ResultsView.tsx
```diff
@@ -156,11 +162,13 @@ const ResultView = ({
   ctrlClick,
   views,
   extraColumn,
+  onRefresh,
 }: {
   r: Result;
   ctrlClick?: (e: Result) => void;
   views: { column: string; mode: string; value: string }[];
   extraColumn?: { row: (e: Result) => React.ReactNode; reserved: RegExp[] };
```
dvargas92495, line 170: similarly, we will one day get rid of `extraColumn` and `ctrlClick` in favor of the user being able to denote `actions` in columns.

src/components/TldrawCanvas.tsx
```diff
@@ -397,51 +394,28 @@ const TldrawCanvas = ({ title }: Props) => {
...
-        userId,
-        uid: persisted.children[0].uid,
-      };
     }
+    const instanceId = TLInstance.createCustomId(pageUid);
+    const userId = TLUser.createCustomId(getCurrentUserUid());
+    const props = getBlockProps(pageUid);
+    const rjsqb = props["roamjs-query-builder"] as Record<string, unknown>;
+    const data = rjsqb?.tldraw as Parameters<TLStore["deserialize"]>[0];
+    return { instanceId, userId, data };
```
dvargas92495, line 418: Instead of a giant text blob, we are now storing the tldraw data in `block/props`. this makes it queryable!

src/components/TldrawCanvas.tsx
```diff
@@ -486,15 +474,59 @@ const TldrawCanvas = ({ title }: Props) => {
...
+  useEffect(() => {
+    const actionListener = ((
+      e: CustomEvent<{
+        action: string;
+        uid: string;
+        val: string;
+        onRefresh: () => void;
+      }>
+    ) => {
+      if (!/canvas/i.test(e.detail.action)) return;
```
dvargas92495, line 495: the way that query builder actions work is that they fire a global event, which our tldraw listens to for any `canvas` type actions

src/index.ts
```diff
@@ -351,6 +363,11 @@ svg.rs-svg-container {
     const toggleDiscourseGraphsMode = await initializeDiscourseGraphsMode(
       onloadArgs
     );
+    if (getNodeEnv() === "development" && localStorageGet(SETTING)) {
+      extensionAPI.settings.set(SETTING, true);
+      toggleDiscourseGraphsMode(true);
```
dvargas92495, line 368: I was getting annoyed at constantly reloading qb and dg being off

src/utils/conditionToDatalog.ts
```diff
@@ -186,6 +209,30 @@ const translator: Record<string, Translator> = {
     placeholder: "Enter any placeholder for the node",
     isVariable: true,
   },
+  "is referenced by block in page with title": {
```
dvargas92495, line 212: this is a mouthful, but the alternative was reworking the `not` condition to be able to handle multiple clauses

src/utils/conditionToDatalog.ts
```diff
@@ -661,6 +711,82 @@ const translator: Record<string, Translator> = {
     },
     placeholder: "Enter any natural language date value",
   },
+  "is in canvas": {
```
dvargas92495, line 714: this is cool. Canvas data is now queryable

src/utils/getBlockProps.ts
```diff
@@ -0,0 +1,22 @@
+import type { PullBlock } from "roamjs-components/types/native";
+
+export const normalizeProps = (
```
dvargas92495, line 3: props come out of Roam with `:` prefixed but go into Roam without it. Go figure

src/components/ResultsView.tsx
mdroidian: RE: `ctrlClick`

It might be a good idea to think about having a quicker way to achieve add to a canvas page or having `action:canvas` on by default on canvas pages.

I imagine if the only way a user could add to canvas via the drawer is by:

- clicking menu
- clicking column views
- clicking view
- clicking action
- typing `canvas`

This could get tedious each time.

src/index.ts
```diff
@@ -36,6 +36,10 @@ import { renderTldrawCanvas } from "./components/TldrawCanvas";
 import getSamePageAPI from "@samepage/external/getSamePageAPI";
 import { registerDatalogTranslator } from "./utils/conditionToDatalog";
 import { QBGlobalRefs } from "./utils/types";
+import localStorageSet from "roamjs-components/util/localStorageSet";
+import localStorageGet from "roamjs-components/util/localStorageGet";
```
mdroidian, line 40: Duplicate?

src/utils/conditionToDatalog.ts
```diff
@@ -76,6 +80,25 @@ const getTitleDatalog = ({
       ];
     }
   }
+  const currentMatch = /^\s*{current}\s*$/i.test(target);
+  if (currentMatch) {
+    // Can't use this, since it's async
+    // window.roamAlphaAPI.ui.mainWindow.getOpenPageOrBlockUid();
+    const uid = getCurrentPageUid();
```
mdroidian, line 87: So this will be the current page at time of query, not the current page of the query itself.
If that is the desired results, cool.
But running this in the sidebar will get the current page results.

mdroidian (approved):
This is 🔥!   Love it.

I added 2 lines to the README, the rest of the changes are just from prettier
- `is in canvas`
- `is referenced by block in page with title`

(I had `{current}` in the wrong spot 😶, so I removed it.  I'll come back to it.)

`{current}` is awesome too, btw.  Just the other day I was going to add an issue for this feature. 🚀

src/index.ts
dvargas92495: Set vs Get

src/utils/conditionToDatalog.ts
dvargas92495: Yea it's meant to be the page of the main window (query drawer doesn't even have a current page). We don't yet have the use case for current page of the query itself

src/components/ResultsView.tsx
dvargas92495: yea... maybe this should be a `Selection` instead of a View type? Would cut out all five of those steps.

The only weird thing would be how it interacts with views, but maybe we could filter it out of that process entirely

src/index.ts
mdroidian: image

src/components/ResultsView.tsx
dvargas92495: Ok I like this alot more. Here's the selection
Screen Shot 2023-05-05 at 2 58 07 PM

And here's the rendering. Notice that we could handle icons now!
Screen Shot 2023-05-05 at 2 58 17 PM

This gets a lot cooler with custom table width setting

PR #74, opened 2023-05-02
Or branches

dvargas92495:
https://www.loom.com/share/7ae0c6fe012547ce89bff0213ef16d83

### DIFF
--- a/src/utils/fireQuery.ts
+++ b/src/utils/fireQuery.ts
@@ -10,6 +10,7 @@ import { DAILY_NOTE_PAGE_REGEX } from "roamjs-components/date/constants";
 import { getNodeEnv } from "roamjs-components/util/env";
 import type { Condition, Result as QueryResult, Selection } from "./types";
 import getSamePageAPI from "@samepage/external/getSamePageAPI";
+import gatherDatalogVariablesFromClause from "./gatherDatalogVariablesFromClause";
 
 export type QueryArgs = {
   returnNode: string;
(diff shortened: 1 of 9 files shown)


src/utils/fireQuery.ts
```diff
@@ -91,35 +92,6 @@ const firstVariable = (
   return undefined;
 };
 
-const getVariables = (
```
dvargas92495, line 94: Moved to `src/utils/gatherDatalogVariablesFromClause.ts`

mdroidian:
`npm i` gave me `npm ERR! ERESOLVE could not resolve`
some googling said to run `npm config set legacy-peer-deps true`
then `npm i` gave me `'patch-package' is not recognized ...`
after `npm i patch-package` and `npm i` I was left with a bunch of `package-lock.json` changes
and `"patch-package": "^7.0.0",` added to `package.json`

then `npm start` cannot find `esbuild`
if I install that, it goes on to another one (`zod`, `dotenv`)

I'm not sure what's going on here.

Same thing if I try a fresh install.

mdroidian:
Does this have to do with upgrading my node version?
I'm having similar issues with `npm start` on smartblocks

mdroidian:
looks like `npm config set legacy-peer-deps true` was the culprit 😣 for 
> then npm start cannot find esbuild
> if I install that, it goes on to another one (zod, dotenv, ...)

`npm i --force` seemed to do the trick.

mdroidian (approved):
Overall this is great!  However, I worry that it might be quite tedious to edit multiple `OR` clauses, deeply nested queries, or get a sense of what the query was without scrolling to the bottom and opening the english view.

Couple ideas to help:
For long results (embedded or more that ~5 or so), a user would have to look at the english view at the bottom, then scroll up and remember which or condition to click into

Possible Solutions: 
1) ability to click onto the english query to jump directly to a branch view to edit
2) an inline english query of selected `OR` condition on hover of `(x) Branches` or maybe via a switch
 
and I love that query icon 🔥

dvargas92495:
Your installation issues are a function of this: https://github.com/dvargas92495/roamjs-query-builder/pull/67/files#r1174180528

In general, I would not `npm config set legacy-peer-deps true`. We want to always tend to be on the latest version, and we heavily use `peer-deps` to install related packages automatically.

> I worry that it might be quite tedious to edit multiple OR clauses, deeply nested queries

I kind of view this as a feature than a bug. `OR`s should be discouraged, especially deeply nested ones.

> ability to click onto the english query to jump directly to a branch view to edit

I love this idea in general. I'm going to look into adding during the style polish tomorrow.

> an inline english query of selected OR condition on hover of (x) Branches or maybe via a switch

I like the english query on hover idea here.

PR #76, opened 2023-05-03
Exploring ideas around custom view types in graph

dvargas92495:
Roam has this concept of Changing Block View:
Screen Shot 2023-05-04 at 12 42 47 PM

This brain blasted me into thinking we should allow users to create their own custom views of data to replace other pieces of UI.

This is one such use case. For now, the component is very specific. But in SamePage, we will want to allow users to create their own UIs instead, similar to smartblocks and queries. Ongoing, SB and QB will continue to serve as testing beds for ideas we want to generalize into SamePage

I wouldn't try to set this up locally (again, it's very custom to a specific user), but would rather review the idea and ask related questions if there are any

### DIFF
--- a/src/components/MessageBlock.tsx
+++ b/src/components/MessageBlock.tsx
@@ -10,11 +10,17 @@ type QueryPageComponent = (props: { blockUid: string }) => JSX.Element;
 
 type Props = Parameters<QueryPageComponent>[0];
 
+/**
+ * This component is hyper custom towards a single user's use case, so we want to keep it mostly hidden for now.
+ * Once Roam releases custom block views, we'll be able to expose a way (via SamePage) to allow users to build
+ * their own custom views (UI) in a No-Code way, similar to SmartBlocks (automations) and Query Builder (Requests) today.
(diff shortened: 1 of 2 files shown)


mdroidian:
So this is like Tana's "Views"
image

Or Notions "Layout"?
image


But allowing user end programming.  Nice.

I'm very interested in something similar for the canvas.  Being able cycle between these views
- Force Directed
- Horizontal Tree
- Vertical Tree
- Manual 1
- Manual 2
- Based on select attributes
  - Heatmap
  - Calendar
  - Bar/Line Chart
  - etc

dvargas92495:
Sort of - except those layout configs are on _queries_, which is more analogous to query builder's layouts.

I'm not sure I've seen a direct parallel to another app yet, in particular allowing end users to build one themselves

mdroidian:
Have you seen https://blockprotocol.org/?  Something similar to that?

dvargas92495:
This is super interesting - there could be some synergy here

PR #82, opened 2023-05-16
Migrating Results Table into its own file

dvargas92495:
Next Steps:
- [x] Sanity check Regressions, particularly the double footer

### DIFF
--- a/src/components/QueryDrawer.tsx
+++ b/src/components/QueryDrawer.tsx
@@ -1,7 +1,6 @@
 import { H3, InputGroup, Button, Tooltip } from "@blueprintjs/core";
 import React, {
   useCallback,
-  useEffect,
   useMemo,
   useRef,
   useState,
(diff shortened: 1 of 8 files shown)


src/components/QueryDrawer.tsx
```diff
@@ -255,9 +253,6 @@ const SavedQuery = ({
         hideResults={minimized}
         results={results.map(({ id, ...a }) => a)}
         preventExport
-        ctrlClick={(r) => {
-          clearOnClick?.(r.text);
-        }}
```
dvargas92495, line 260: replaced with `-action` selections

src/components/ResultsTable.tsx
```diff
@@ -0,0 +1,526 @@
...
+  setPage: (p: number) => void;
+  onRefresh: () => void;
+  allResultsLength: number;
+  // @deprecated - move to `-action` columns through selections
+  extraColumn?: {
+    reserved: RegExp[];
+    width: number;
+    row: (e: Result) => React.ReactNode;
+    header: React.ReactNode;
+  };
```
dvargas92495, line 332: I think over time, we need to decide which of these settings are actually for all layouts, and which ones are specific to table

src/components/ResultsView.tsx
```diff
@@ -1173,41 +698,42 @@ const ResultsView: ResultsViewComponent = ({
...
-                      Showing {paginatedResults.length} of {results.length}{" "}
-                      results
-                    </i>
-                  </span>
-                </div>
-                {showContent && <QueryUsed parentUid={parentUid} />}
-              </div>
             </>
           ))}
+        <div style={{ background: "#eeeeee80" }}>
```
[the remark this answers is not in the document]
mdroidian: My suggestion would be to add 4px padding to bring it more in line with the rest.
`ResultsView`
`padding: 0 4px` on ln 716

And I think that was the padding for the ResultsTable footer.
`ResultsTable`
`padding: 8px 4px` on ln 452
`padding-left: 7px` on ln 453

image
image

mdroidian (changes requested):
Requested changes in comments

PR #85, opened 2023-05-18
Automatically draw existing edges

dvargas92495:
https://www.loom.com/share/09de0f47b802405fbd6f84992af7d60d


dvargas92495:
ah I missed that case as well! Will include in my iteration /tomorrow

PR #87, opened 2023-05-22
Adding Page as a Default Node configured

dvargas92495:


src/utils/getDiscourseNodes.ts
```diff
@@ -3,11 +3,64 @@ import getSubTree from "roamjs-components/util/getSubTree";
...
+  canvasSettings: {
+    [k: string]: string;
+  };
+  // @deprecated - use specification instead
+  format: string;
+};
 
-const getDiscourseNodes = (relations = getDiscourseRelations()) =>
-  Object.entries(discourseConfigRef.nodes)
+const DEFAULT_NODES: DiscourseNode[] = [
```
dvargas92495, line 25: This is the biggest idea communicated in the PR - we are now adding default discourse nodes representing Block and Page. This brings all sorts of benefits, including allowing users to include Roam's (and in the future, any app's) base primitives into their discourse schema.

For videos, checkout out the `teamcovid` graph, since their graph appears in the video

dvargas92495:
Known issues to clean up with any feedback:
- Discourse nodes that are blocks are page referencing during relation creation instead of block references

mdroidian (approved):
Minor UI/UX suggestion:
A user can create a Page or DG Node from the canvas.
A user cannot create a Block from the canvas, but the UI is the same.

We should either create the block or change the UI to reflect that a block cannot be created.

https://github.com/dvargas92495/roamjs-query-builder/assets/3792666/540901ac-a137-4953-ad28-3840a60b1d10

dvargas92495:
Ah it's meant to be supported: https://github.com/dvargas92495/roamjs-query-builder/pull/87/files#diff-2a2ebc962f7a0b29a81264b90b97a81f096b2ac88c73aabcd866a921f3a69b3cR129-R138 - will debug this as well tm

PR #71, opened 2023-04-27
Begin Kanban View

dvargas92495:

PR #89, opened 2023-05-31
add minimal view to ResultsTable

mdroidian:


src/components/ResultsView.tsx
```diff
@@ -286,10 +293,17 @@ const ResultsView: ResultsViewComponent = ({
...
   const debounceRef = useRef(0);
+  const minimal = layout.mode === "minimal" ? true : false;
+  const [showInterface, setShowInterface] = useState(minimal ? false : true);
+  const [showIcons, setShowIcons] = useState(minimal ? false : true);
+  const appear = useCallback(() => setShowIcons(true), [setShowIcons]);
+  const disappear = useCallback(() => setShowIcons(false), [setShowIcons]);
   return (
     <div
-      className="roamjs-query-results-view w-full relative"
+      className={`roamjs-query-results-view w-full relative mode-${layout.mode}`}
```
mdroidian, line 303: For users to add their own styles if they choose.

src/components/ResultsView.tsx
```diff
@@ -286,10 +293,17 @@ const ResultsView: ResultsViewComponent = ({
       );
   };
   const debounceRef = useRef(0);
+  const minimal = layout.mode === "minimal" ? true : false;
+  const [showInterface, setShowInterface] = useState(minimal ? false : true);
+  const [showIcons, setShowIcons] = useState(minimal ? false : true);
+  const appear = useCallback(() => setShowIcons(true), [setShowIcons]);
+  const disappear = useCallback(() => setShowIcons(false), [setShowIcons]);
```
dvargas92495, line 300: In general, we want to start moving state that is specific to only a single layout into its layout comonent. For now, this is okay and can be revisited in the future

src/components/ResultsView.tsx
```diff
@@ -662,6 +691,8 @@ const ResultsView: ResultsViewComponent = ({
                 pageSizeTimeoutRef={pageSizeTimeoutRef}
                 onRefresh={onRefresh}
                 allResultsLength={allResults.length}
+                minimal={layout.mode === "minimal"}
+                showInterface={showInterface}
```
dvargas92495, line 695: This makes me think that this could be layout settings for `table` instead of a separate layout

dvargas92495 (commented):
Some thoughts on organinzation

src/components/ResultsView.tsx
mdroidian: While I'm hesitant to add another menu item for a single layout and a single setting, I agree with you, as it will allow for further settings per layout in the future 👍

src/components/ResultsView.tsx
mdroidian: Would that then entail moving the controls to each layout?

src/components/ResultsView.tsx
dvargas92495: The concept of layout settings I'm starting in the kanban PR #71, taking inspiration for how Notion does layout settings which works rather well

src/components/ResultsView.tsx
```diff
@@ -293,17 +293,20 @@ const ResultsView: ResultsViewComponent = ({
       );
   };
   const debounceRef = useRef(0);
-  const minimal = layout.mode === "minimal" ? true : false;
-  const [showInterface, setShowInterface] = useState(minimal ? false : true);
-  const [showIcons, setShowIcons] = useState(minimal ? false : true);
+  const [showInterface, setShowInterface] = useState(
+    layout.mode === "minimal" ? false : true
```
dvargas92495, line 297: same as `layout.mode !== "minimal"`

mdroidian:
clear layout settings

src/components/ResultsView.tsx
```diff
@@ -441,13 +498,62 @@ const ResultsView: ResultsViewComponent = ({
...
+                              });
+                              setInputSetting({
+                                key: s.key,
+                                value: (e.target as HTMLInputElement).checked
+                                  ? s.options[0]
+                                  : "default",
+                                blockUid: layoutNode.uid,
+                              });
+                              setShowInterface(
+                                layout.style !== "Minimal" &&
```
mdroidian, line 546: Not a big fan of this

dvargas92495: The conflict here appears to be doing something that feels like it's layout specific but is actually resultsview-wide.

I think a decision should be made here: should a interface toggle be layout specific or global? then commit to the decision

mdroidian: I went with global

src/components/ResultsView.tsx
```diff
@@ -546,6 +613,27 @@ const ResultsView: ResultsViewComponent = ({
...
+                    text={showInterface ? "Hide Interface" : "Show Interface"}
+                    onClick={() => {
+                      const resultNode = getSubTree({
+                        key: "results",
+                        parentUid,
+                      });
+                      const layoutNode = getSubTree({
+                        key: "layout",
+                        parentUid: resultNode.uid,
+                      });
```
dvargas92495, line 627: IIUC, this is only being used in table correct? If so, it could be treated similarly to how we are treating `Row Style`. If not, meaning we foresee using it in other layouts, then let's remove it from the layout tree and add it to the results tree

src/components/ResultsView.tsx
```diff
@@ -286,11 +314,20 @@ const ResultsView: ResultsViewComponent = ({
       );
   };
   const debounceRef = useRef(0);
+  const [showInterface, setShowInterface] = useState(
+    layout.interface !== "hide"
+  );
+  const [showMenuIcons, setShowMenuIcons] = useState(
+    layout.interface !== "hide"
```
dvargas92495, line 321: both showMenIcons and setInterface are being source from the same persistent state, which either means it should be stored separately or we have redundant local state

mdroidian: Sorry, I don't understand.

Are you suggesting we should store `layout.interface` in a `useState` hook as well??

Both `showMenuIcons` and `showInterface` use the `layout.interface` for their initial state, but their states change depending on different events.  (Eg, `showMenuIcons` updates on hover)

src/components/ResultsView.tsx
mdroidian: It currently applies to all layouts.

https://github.com/dvargas92495/roamjs-query-builder/assets/3792666/4709bf83-3859-4a4b-aedf-0e3b5983a2c6

src/components/ResultsView.tsx
dvargas92495: In which case, it seems like `showMenuIcons` is actually not dependent on the value of `layout.interface`, since we check them both independently on `L378`, so it should probably just be initialized as `useState(false)`.

This is non-blocking feedback though, will continue to merge

src/components/ResultsView.tsx
```diff
@@ -546,6 +613,23 @@ const ResultsView: ResultsViewComponent = ({
...
+                        key: "results",
+                        parentUid,
+                      });
+                      setInputSetting({
+                        key: "interface",
+                        value: showInterface ? "hide" : "show",
+                        blockUid: resultNode.uid,
+                      });
+                      setShowInterface((s) => !s);
+                      setMoreMenuOpen(false);
```
dvargas92495, line 630: oh wait, one last thing before we merge - now that this is sourced from `results.interface` we need to update the `useState` for `showInterface` to source from the correct block instead of layout

mdroidian: Nice catch

PR #92, opened 2023-06-06
Autosizing discourse nodes based on text

dvargas92495:


src/components/TldrawCanvas.tsx
```diff
@@ -479,7 +482,8 @@ class DiscourseNodeUtil extends TLBoxUtil<DiscourseNodeShape> {
       if (
         shape.props.uid !== loaded &&
         !isPageUid(shape.props.uid) &&
-        contentRef.current
+        contentRef.current &&
+        isLiveBlock(shape.props.uid)
```
dvargas92495, line 486: fixes a bug related to rendering non existent blocks

mdroidian (commented):
I found the padding to be inconsistent.  As a user, this felt jarring. 

Especially the difference between the longest three tests, which would likely be more typically of actual user input on a discourse graph.

It seems that the longer the text, the greater the top and bottom padding.

image

PR #95, opened 2023-06-09
Two way syncing in TLDraw - single graph

dvargas92495:


src/components/TldrawCanvas.tsx
```diff
@@ -945,13 +1017,19 @@ const TldrawCanvas = ({ title }: Props) => {
...
       clearTimeout(serializeRef.current);
-      serializeRef.current = window.setTimeout(() => {
+      serializeRef.current = window.setTimeout(async () => {
         const state = _store.serialize();
         const props = getBlockProps(pageUid) as Record<string, unknown>;
         const rjsqb =
           typeof props["roamjs-query-builder"] === "object"
             ? props["roamjs-query-builder"]
             : {};
+        // we need this bc Roam doesn't update edit/user or edit/time when we just edit block/props
```
dvargas92495, line 1027: Josh has acknowledged this as a bug

mdroidian: Looks like `edit/time` is updating now.  But not `edit/user`.
Tested and ref: https://roamresearch.com/#/app/help/page/O3JfglQsT

mdroidian (approved):
Looks good.  Interesting "side effect" that the last mouse position of the user who initiated the last change is shown ... but it seems to disappear after ~10 seconds.


https://github.com/dvargas92495/roamjs-query-builder/assets/3792666/bae4fbae-e25e-45fb-98c9-6f2e82e6fa41

PR #103, opened 2023-06-22
set Condition/Selection Inputs onBlur w/o timeout

mdroidian:
What about something like this?

If acceptable, I'll update for:
 - condition target
 - selection label
 - selection value

### DIFF
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -85,23 +85,17 @@ const QueryClause = ({
       setConditions((_conditions) =>
         _conditions.map((c) => (c.uid === con.uid ? { ...c, relation: e } : c))
       );
-      if (timeout) {
-        debounceRef.current = window.setTimeout(() => {
+      debounceRef.current = window.setTimeout(
+        () => {
           setInputSetting({
             blockUid: con.uid,
             key: "Relation",
             value: e,
             index: 1,
           });
-        }, 1000);
-      } else {
-        setInputSetting({
-          blockUid: con.uid,
-          key: "Relation",
-          value: e,
-          index: 1,
-        });
-      }
+        },
+        timeout ? 1000 : 0
+      );
     },
     [setConditions, con.uid]
   );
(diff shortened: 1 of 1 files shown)


src/components/QueryEditor.tsx
```diff
@@ -143,6 +146,7 @@ const QueryClause = ({
         <AutocompleteInput
           value={con.relation}
           setValue={setConditionRelation}
+          onBlur={(e) => setConditionRelation(e, false)}
```
dvargas92495, line 149: Id be ok with calling `setConditionRelation` directly instead of the other usage, then not needing the timeout param

mdroidian: Do you mean dropping `setValue` and the timeout altogether?

dvargas92495: ooh hm. okay let's see how keeping both fares for now

mdroidian:
Interesting.  It is showing your last commit as changes on this pr (samepage version and fireQuery.ts)

I did 
`git checkout main`
`git pull`
`git checkout update-input-onblur`
`git rebase main`

src/components/QueryEditor.tsx
```diff
@@ -156,6 +159,7 @@ const QueryClause = ({
         <AutocompleteInput
           value={con.target}
           setValue={setConditionTarget}
+          onBlur={(e) => setConditionTarget(e, false)}
```
dvargas92495, line 162: shouldn't this first argument be `con.target` instead of `e` ?

dvargas92495: the other two seem to be sending an event so those should be fine

dvargas92495:
I think the `pull` is where things may have gotten wonky, and specifying `git pull origin main` would've pulled just the main branch.

Once the last piece of feedback is addressed, I'll pull the changes in a separate branch and merge into main

src/components/QueryEditor.tsx
mdroidian: `setValue` seems to set `con.target` for an autocomplete on hover, even if the user doesn't select it.  So `e` looks to be a more accurate reflection of what the user experiences.

https://github.com/dvargas92495/roamjs-query-builder/assets/3792666/7e818ce3-ddda-4fef-8ef3-42313860765f

PR #28, opened 2023-01-31
Update to docs - add details/summary

mdroidian:
added
`<details>` `<summary>` in a few places

Tested manually in `index.mdx` in `roamjs-com`, seemed good.

https://user-images.githubusercontent.com/3792666/215666689-3d3f36a2-fb1d-4101-beac-863e88bca109.mp4

@dvargas92495

PR #104, opened 2023-06-26
add examples, table, details/summary

mdroidian:
Requested from [customer](https://roamresearch.com/#/app/teamcovid/page/LJMFrs38_) and follow up on #28.

More examples to be added in future PR's.
Could start adding query templates in `/EXAMPLES.md` as well

Unsure if `/EXAMPLES.md` in root or if we should add something like `/docs`


mdroidian:
Ah, I just realized roamjs.com probably won't support an additional `.md` file.  And Roam still doesn't support the relative links from the Table of Contents.  I wonder if we should just move all of the docs to `/docs` and direct Roam Depot and roamjs.com to GitHub.

- Root readme directs to `/docs/README.md`
- Move `README.md` to `/docs/README.md`
- Move `EXAMPLES.md` to `/docs/EXAMPLES.md`

dvargas92495:
it's easy to update roamjs.com to whatever Roam supports. The question is what does Roam support? Since it seems like they don't support relative paths, we probably want to update each relative path to be an absolute path to roamjs.com/extensions/query-builder/X instead, and then have RoamJS render that markdown file from github

mdroidian:
If we redirect all of the relative paths (I'm guessing that includes the ToC) then GitHub ToC will also redirect to roamjs.com?

It seems like a very confusing UX having majority of the docs content on a site (Roam / GitHub) but being redirected to a different site clicking on the ToC or example links.

Would it be more efficient to pick one main source (GitHub or roamjs) for the docs and direct everyone there?  That would also save time triple checking each addition for compatibility.

dvargas92495:
yea so the pattern with most `npm` libraries is to just redirect to their own site, providing minimal examples in the README, which was the original intent with roamjs.com (see SmartBlocks docs). I think that's where I'd lean as well, but open to any approach you feel is more intuitive

mdroidian:
I see the benefit of self hosting if we had something like a single product, thousands of MAU's, or fulltime staff. But with our current state and foreseeable future, I lean towards GitHub with the same format as SmartBlocks / WorkBench, except the external links are to github.com/repo/docs:

README.md
 - ToC
   - section 1 (external link to 1)
   - section 2 (external link to 1)
   - ...


It seems we have at least 3 options:
1. host docs in a Roam graph, custom render on roamjs (SmartBlocks, WorkBench)
2. host docs in GitHub, custom render on roamjs
3. host docs in GitHub

If there were immediate plans to build out the RoamJS Static Site, then option 1 would make sense.
At the moment, I don't see the added value of option 2, but I do see overhead to feature parity (search, static ToC, formatting, etc) and additional maintenance (API, hosting, regenerating when changes are made, etc)

With GitHub it seems fairly straightforward, easier developer contribution, and we can focus on shipping features / fixing bugs.

While I lean GitHub, I'm happy to go in either direction 👍

mdroidian:
I did an `npm version patch` before I remembered that you did one earlier today.  So I just did another one and used GitHubs UI to merge the conflict.

dvargas92495 (approved):
This should get a `npm version patch` too. Feel free to merge once that's up

mdroidian:
re: `npm version patch`

Sorry, just wanted to clarify.  I did bump the version from 1.8.14 to 1.8.15.

Did you mean `npm version minor`?

PR #109, opened 2023-06-29
Add condition examples

mdroidian:
https://github.com/dvargas92495/roamjs-query-builder/blob/docs-add-examples/docs/examples.md
or a screenshot: https://github.com/dvargas92495/roamjs-query-builder/assets/3792666/4587deba-2d7c-459d-8c82-d6aed330799c

### DIFF
--- a/src/components/ResultsView.tsx
+++ b/src/components/ResultsView.tsx
@@ -386,6 +386,7 @@ const ResultsView: ResultsViewComponent = ({
             </Tooltip>
           )}
           <Popover
+            placement="left-end"
             isOpen={moreMenuOpen}
             target={
               <Button
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -2,59 +2,92 @@
 
 ## Created After
 
-Coming Soon.
+`created after` - The `source` block or page was created after the naturally specified `target`
+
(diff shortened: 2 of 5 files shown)


src/components/ResultsView.tsx
```diff
@@ -386,6 +386,7 @@ const ResultsView: ResultsViewComponent = ({
             </Tooltip>
           )}
           <Popover
+            placement="left-end"
```
mdroidian, line 389: I also snuck this in here to deal with the placement of `Edit Query` being unclickable

image

dvargas92495: how does this behave if the qb is at the bottom of the page? I guess you could always scroll down further

dvargas92495 (approved):
Great stuff here 🔥

src/components/ResultsView.tsx
mdroidian: In my tests, worst case scenario is `Copy Query` is hidden when there are no results.  But a user can also toggle footer to alleviate that.

https://github.com/dvargas92495/roamjs-query-builder/assets/3792666/8e405057-63a8-4064-b1c0-7e4c6b351f6e

PR #115, opened 2023-07-05
Selection suggestions

dvargas92495:


mdroidian (changes requested):
Noticed a few things:

- node:(created by, created since, etc) isn't valid
  - I believe it's either `node:node:created by` or just `created by` which are both available
- I suggest we remove duplicates (Author/Created By, edit/edited)
  - As a user, if I saw `edit ...` and `edited ...` I would think that they were both there because they had different functions.  I believe this just adds confusion
  - it's fine to leave them valid selections, but no need to suggest both
- all pages show up instead of just `Attribute::`s
  - Are they used at all?  If not, adds a lot more noise than required
- add/subtract replace `node` with variables.  But I think they uses the selection label instead of condition variables

Otherwise, this is 🔥

dvargas92495:
- Agreed
- Agreed
- Quasi-agree - the problem here is I _think_ a query to get just the page titles used as attributes would be expensive, but I could test this thesis on revision
- Agreed

mdroidian:
FYI, I just tried this in teamcovid graph (large graph) and it was extremely sluggish on first keypress.

Maybe it was my computer, but it's worth testing on yours.

https://github.com/RoamJS/query-builder/assets/3792666/df07a5a1-0905-490f-a558-3b929996d4a6

mdroidian:
Another one: `action`

`action:node:canvas` shows up, but it should be `action:canvas:{button}` (eg: `action:canvas:plus`)

dvargas92495:
worth testing on that graph

today was hectic so didn't get around to feedback - let's tackle together during pair programming

PR #118, opened 2023-07-10
Filter Menu Item

dvargas92495:


mdroidian (commented):
### UI Suggestions

**Add transparent border** to inner elements or padding to container
image
image

**Add notification** that filter is applied.  
For search, I made the menu and line item red (which also matches with the roam style filters):


https://github.com/RoamJS/query-builder/blob/541981068b9bf9a2c4fd8fb0bde3f2146ba4cf73/src/components/ResultsView.tsx#L392-L396

https://github.com/RoamJS/query-builder/blob/541981068b9bf9a2c4fd8fb0bde3f2146ba4cf73/src/components/ResultsView.tsx#L692-L696

image

image


**Move Search** To Under Filters
I put it in wrong spot 😁
It would go with the other "Edit Results View" options; Layout, Filters, Column Views, Hide Interface.
Where Edit Block, Delete Query, and Copy Query belong grouped together.  Actually I'd move Delete to the bottom as well.  This could probably be done in another UI related PR.

image

PR #119, opened 2023-07-11
Simplify interfaces

dvargas92495:


mdroidian:
For some reason using `enter` on `Open Query Drawer` causes it to open and close.

https://github.com/RoamJS/query-builder/assets/3792666/9d5fe5ad-40d4-4fc2-8bec-ebd42127a83c

The following are not directly a result of these changes, but I added them here because they are related to `node`

1) `node` discourse relations don't support autocomplete on target whereas `Claim` (or any valid discourse node) did.

image
image

2) `Has Any Relation To` doesn't work with `node`

mdroidian:
RE: discourse relations as condition relationship

The autocomplete looks good, but the actual query doesn't seem to work.

I tried `Supports` `Evidence` as well as `Supports` `{{individual node}}`
I don't think this is from these changes though.  It doesn't work in the live version either.
It seems the only way to get it to work is from changing the return `node` to `Evidence`

I believe that is what this `is a` [comment](https://roamresearch.com/#/app/teamcovid/page/6M7u8gwdw) is referring to

PR #120, opened 2023-07-12
ATTR_SUGGESTIONS hangs graph fix

dvargas92495:
Took over 10 min to run on one client's graph, causing it to hang.

Going to merge right away to fix the hang


mdroidian:
👍

PR #125, opened 2023-07-16
Centralize the Export/ExportDialog components in preparation for #113

dvargas92495:
No need to test this yet since there's going to be a followup with the actual exporting logic. This is more of a refactor FYI


src/components/CytoscapePlayground.tsx
```diff
@@ -67,6 +69,69 @@ const editCursor =
 const trashCursor =
   "https://raw.githubusercontent.com/dvargas92495/roamjs-discourse-graph/main/src/cursors/trash.png";
 
+type ExportDialogProps = {
```
dvargas92495, line 72: This is a copy paste of the old `ExportDialog` component. will be nuked once we nuke Cytoscape since this was its last usage

src/components/Export.tsx
```diff
@@ -289,4 +294,7 @@ export const ExportDialog: ExportDialogComponent = ({
   );
 };
 
+export const render = (props: ExportDialogProps) =>
+  renderOverlay({ Overlay: ExportDialog, props });
+
 export default ExportDialog;
```
dvargas92495, line 300: In the future, would probably make sense to rename this file to `ExportDialog`. Didn't do it in this PR to limit confusion

src/discourseGraphsMode.ts
```diff
@@ -748,7 +750,36 @@ const initializeDiscourseGraphsMode = async (args: OnloadArgs) => {
...
+                  isSamePageEnabled,
+                })
+              )
+            ).then((r) => r.flat());
+          exportRender({
+            results,
+            exportTypes: getExportTypes({
+              results,
+            }),
+          });
```
dvargas92495, line 781: API could probably be simplified further

dvargas92495: Also, from this point forward we should look to move logic out of `discourseGraphsMode` to back within the QueryBuilder extension at large. We then would want to nuke this option, and the Discourse Graph grammar is a downloadable schema from the grammar registry we talked about

PR #133, opened 2023-08-04
add "send to canvas" on export

mdroidian:
- [x] Send To Canvas, if current page = canvas (via `roamjs:query-builder:action`)
- [x] Send To Page, if current page = page (via `clearOnClick`)
- [x] Add ability to send to specific canvas
- [x] Add ability to send to specific page
- [-] Remove "Insert Results" from header

Customer wanted to keep it there, so instead
- [x] "Insert Results" opens Export dialog

What's left
- [x] `addToSelectedCanvas`: detect what type of node the result is, add that type to canvas
- [x] `addToSelectedPage`: detect what type of node the result is, if block, add ref to page
- [x] support dnp as current (actually, this might work already: test it)
- [x] remove duplicate `<Export>` in `QueryDrawer.tsx`
- [x] remove console logs / clean up

Also, blueprint is handling setting shown tabs/panel, might need to make that controlled by us because `Export Discourse Graph` shows `Send To` as first panel.

### DIFF
--- a/src/components/Export.tsx
+++ b/src/components/Export.tsx
@@ -33,6 +33,7 @@ import { defaultDiscourseNodeShapeProps } from "./TldrawCanvas";
 import AutocompleteInput from "roamjs-components/components/AutocompleteInput";
 import getAllPageNames from "roamjs-components/queries/getAllPageNames";
 import getPageUidByPageTitle from "roamjs-components/queries/getPageUidByPageTitle";
+import findDiscourseNode from "../utils/findDiscourseNode";
 
 const ExportProgress = ({ id }: { id: string }) => {
   const [progress, setProgress] = useState(0);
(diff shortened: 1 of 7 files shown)


src/components/Export.tsx
```diff
@@ -100,157 +107,202 @@ const ExportDialog: ExportDialogComponent = ({
...
+        isCloseButtonShown={false}
+        title={`Export Query Results`}
+        autoFocus={false}
+        enforceFocus={false}
+        portalClassName={"roamjs-export-dialog-body"}
+      >
+        <Tabs id="export-tabs" large={true}>
+          <Tab id="sendto" title="Send To" panel={SendToPanel} />
+          <Tab id="export" title="Export" panel={ExportPanel} />
+        </Tabs>
```
mdroidian, line 305: This is the most relevant change in `Export.tsx`, along with passing `parentUid` to export

Most of the add/delete is from moving ExportPanel into its own component.

src/components/Export.tsx
```diff
@@ -100,157 +113,230 @@ const ExportDialog: ExportDialogComponent = ({
...
+            style={{ minWidth: 64 }}
+            disabled={loading}
+          />
+        </div>
+      </div>
+    </>
+  );
+
+  const currentPageUid = getCurrentPageUid();
+  const currentPageTitle = getPageTitleByPageUid(currentPageUid);
```
mdroidian, line 262: didn't combine these because I think we'll I'll be using `currentPageUid` later

src/components/Export.tsx
```diff
@@ -100,157 +113,230 @@ const ExportDialog: ExportDialogComponent = ({
...
+        </span>
+      </div>
+      <div className={Classes.DIALOG_FOOTER}>
+        <div className={Classes.DIALOG_FOOTER_ACTIONS}>
+          <Button text={"Cancel"} intent={Intent.NONE} onClick={onClose} />
+          <Button
+            text={"Send"}
+            intent={Intent.PRIMARY}
+            onClick={() => {
+              console.log(`isCanvasPage`, isCanvasPage);
```
mdroidian, line 291: remove this

src/components/QueryDrawer.tsx
```diff
@@ -75,6 +77,13 @@ const SavedQuery = ({
         margin: 4,
       }}
     >
+      <Export
+        isOpen={isExportOpen}
+        onClose={() => setIsExportOpen(false)}
+        results={results.map(({ id, ...a }) => a)}
+        parentUid={uid}
+        clearOnClick={clearOnClick}
+      />
```
mdroidian, line 86: This seems inefficient.  How can I get the state with line 134 to open the `Export` in `Results View`

mdroidian:
tagging @dvargas92495 for awareness.  

Should be cleaned up and ready to merge tomorrow.  Will probably merge Wednesday if not reviewed tomorrow.

mdroidian:
@dvargas92495 in addition to what's mentioned in the video, also nuked a bunch of the `clearOnClick`

https://www.loom.com/share/c954ffd0f79b4836a8ee9b14c6dc2f86

PR #140, opened 2023-08-10
add check for shape title, set editingId

mdroidian:


src/components/TldrawCanvas.tsx
```diff
@@ -631,6 +631,11 @@ class DiscourseNodeUtil extends TLBoxUtil<DiscourseNodeShape> {
       [this.app, shape.id]
     );
     const [isLabelEditOpen, setIsEditLabelOpen] = useState(false);
+    useEffect(() => {
+      if (!shape.props.title) {
+        this.app.setEditingId(shape.id);
+      }
+    }, [shape.props.title, shape.id]);
```
mdroidian, line 638: @dvargas92495 do you remember the reason this code was removed from #122 ?

I believe it is causing the clicking issue when trying to add a node:

https://github.com/RoamJS/query-builder/assets/3792666/1d117471-c91b-453a-945f-57aa2ed46fa5

dvargas92495: I think I just wanted to change it to be aligned with the tldraw editing state, to automatically handle things like double click to edit and any other instance that tldraw sets editing to true

PR #149, opened 2023-08-31
set black / off black for page/block canvas nodes

mdroidian:


src/components/TldrawCanvas.tsx
```diff
@@ -450,7 +450,7 @@ const COLOR_PALETTE: Record<string, string> = {
   white: "#ffffff",
   yellow: "#ffc078",
 };
-const COLOR_ARRAY = Array.from(TL_COLOR_TYPES);
+const COLOR_ARRAY = Array.from(TL_COLOR_TYPES).reverse();
```
mdroidian, line 453: reversed so that the initial custom discourse nodes created by users do not default to black/grey

PR #159, opened 2023-09-25
add disabled/maxItems to canvas labelDialog

mdroidian:
requires https://github.com/RoamJS/roamjs-components/pull/18

Improved UX / perf for `page`, `block`, and discourse nodes with large results


dvargas92495 (approved):
Id use 100

PR #165, opened 2023-10-10
Selective dgraph node export

mdroidian:
https://github.com/RoamJS/query-builder/assets/3792666/7a307d95-ac16-4d13-9fd4-ad7e086df16f
- [x] export via graph overview by discourse context depth
- [x] export via graph overview by any page (degrees in, degrees out)
- [x] explainer text
- [x] clean up code


src/components/ExportDiscourseContext.tsx
```diff
@@ -0,0 +1,393 @@
...
+
+const GraphExportDialog: GraphExportDialogComponent = ({
+  isOpen,
+  onClose,
+  initialDegreesIn = 0,
+  initialDegreesOut = 0,
+  initialDiscourseContextDepth,
+}) => {
+  const [loading, setLoading] = useState(false);
+  const pages = window.roamAlphaAPI.ui.graphView.wholeGraph.getExplorePages();
```
dvargas92495, line 204: memo

src/components/ExportDiscourseContext.tsx
```diff
@@ -0,0 +1,393 @@
...
+  const initialUids = pages.map((x) => getPageUidByPageTitle(x)) || [];
```
dvargas92495, line 205: memo

src/components/ExportDiscourseContext.tsx
```diff
@@ -0,0 +1,393 @@
...
+
+const getReferencesByDegree = async (
+  initialUids: string[],
+  degrees: number,
+  getReferences: (uid: string) => Promise<string[]>
+): Promise<Set<string>> => {
+  let currentUids = [...initialUids];
+  const visitedUids = new Set<string>();
+
+  for (let i = 0; i < degrees; i++) {
```
dvargas92495, line 52: See if we could avoid this loop

`[:find [pull ?b [:block/uid :block/string {:block/refs 3}]] :where [?b :block/uid "dq4_gLsOG"]]`

src/components/ExportDiscourseContext.tsx
```diff
@@ -0,0 +1,393 @@
...
+    const nextUids = [];
```
dvargas92495, line 53: I think this should be a `Set`

dvargas92495 (approved):

src/components/ExportDiscourseContext.tsx
mdroidian: Not super happy with where I got so far, so I'm going to defer this to later.
Also, I don't see a way to query all of the inbound references without a loop.

Here's outbound so far
``` typescript
const getOutBoundReferencesQuery = async (uid: string, depth: number) => {
  //TODO run this for each initial uid
  const adjustedDept = depth + 1
  const query = window.roamAlphaAPI.q(`
    [:find (pull ?parent 
            [:block/uid 
            :node/title
            {:block/children ${adjustedDept} 
              :block/refs ${adjustedDept}}]) 
    :where [?parent :block/uid "${uid}"]]
  `)[0][0] as []

  const extractUids = (node: any[]) => {
    let uids: string[] = [];

    function traverse(node: any) {
      if (Array.isArray(node)) {
        node.forEach(innerNode => traverse(innerNode));
      } else if (typeof node === 'object' && node !== null) {
        if ('uid' in node && 'title' in node) {
          uids.push(node.uid);
        }

        Object.keys(node).forEach(key => {
          if (typeof node[key] === 'object') {
            traverse(node[key]);
          }
        });
      }
    }

    traverse(node);
    return uids;
  }
  const uids = extractUids(query)
  return uids
}
```

PR #168, opened 2023-10-18
Add progress bars, separate preview/export

mdroidian:


src/components/ExportDiscourseContext.tsx
```diff
@@ -117,7 +134,10 @@ const getDiscourseContextResultsByDepth = async ({
         currentDepth: currentDepth + 1,
         maxDepth: maxDepth,
         visitedUids,
+        updateProgressWithDelay,
+        progressRefresh,
       });
+      progressRefresh = progressRefresh === 0.9 ? 0.1 : progressRefresh + 0.1;
```
mdroidian, line 140: Based on this structure I couldn't find a 'meaning' progress segment.
This at least refreshes the UI (to stop the hang) and notifies the user that "something is happening"

Other option would be to rewrite and query by breadth first.

src/components/ExportDiscourseContext.tsx
```diff
@@ -263,54 +323,156 @@ const GraphExportDialog: GraphExportDialogComponent = ({
...
+    if (isDiscourse) {
+      setDiscourseProgress(progress);
+      if (message) setDiscourseProgressMessage(message);
+      return;
+    } else {
+      setProgress(progress);
+      if (message) setProgressMessage(message);
+    }
+    await delay(delayAmount);
+  };
```
mdroidian, line 401: I'm all ears for a better implementation 😁 @dvargas92495

PR #178, opened 2023-11-22
addCommand "Create Query Block" and "Preview" kb shortcut

mdroidian:
https://www.loom.com/share/fa9bb926fda44dff8e4700c700cc5e18

### DIFF
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -33,7 +33,6 @@ import getSubTree from "roamjs-components/util/getSubTree";
 import AutocompleteInput from "roamjs-components/components/AutocompleteInput";
 import getNthChildUidByBlockUid from "roamjs-components/queries/getNthChildUidByBlockUid";
 import getChildrenLengthByPageUid from "roamjs-components/queries/getChildrenLengthByPageUid";
-import getBlockUidFromTarget from "roamjs-components/dom/getBlockUidFromTarget";
 import parseQuery, { DEFAULT_RETURN_NODE } from "../utils/parseQuery";
 import { getDatalogQuery } from "../utils/fireQuery";
 import getTextByBlockUid from "roamjs-components/queries/getTextByBlockUid";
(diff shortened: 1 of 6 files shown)


src/components/QueryPage.tsx
```diff
@@ -159,6 +155,10 @@ const QueryPage = ({
                 setHasResults(true);
                 setIsEdit(false);
               }}
+              setHasResults={() => {
+                setHasResults(true);
+                onRefresh();
```
mdroidian, line 160: I'd be interested in discussing whether it's worthwhile to extend this via something like

- `useContext`
- `window.roamjs.extension.queryBuilder`
- or another way

so that we could use `addCommand` instead of the `EventListener` in QueryEditor

src/components/QueryEditor.tsx
```diff
@@ -436,16 +437,37 @@ const getConditionByUid = (
...
   showAlias,
 }) => {
+  const previewQuery = useCallback(
+    (e) => {
+      if (e.altKey && e.key === "q") {
+        const target = document.activeElement as HTMLElement;
+        const uid = getBlockUidFromTarget(target);
+        if (uid !== parentUid) return;
+        if (setHasResults) setHasResults();
+      }
```
dvargas92495, line 459: - `commandPalette.addCommand()` in `index.ts` so we only set it once (can keep the keyboard shortcut as a default).
- On trigger, `const target = document.activeElement as HTMLElement;const uid = getBlockUidFromTarget(target);`
- end it by calling `document.body.dispatch(new CustomEvent("roamjs-query-builder:fire-query", {detail: uid}))`
- the use effect below is replaces `window.addEventListener` with `document.body.addEventListener("roamjs-query-builder:fire-query")` - `if (uid !== parentUid) return;`

src/index.ts
```diff
@@ -606,6 +606,60 @@ svg.rs-svg-container {
...
+              ],
+            },
+          ],
+        },
+        parentUid: uid,
+      });
+      document.querySelector("body")?.click();
+      setTimeout(() => {
+        const conditionEl = document.querySelector(
+          ".roamjs-query-condition-relation"
```
dvargas92495, line 654: this won't work with multiple query blocks in the same page

dvargas92495 (approved):

PR #179, opened 2023-11-23
Kanban embed view

mdroidian:
PR for client testing.  Waiting on roamAlphaAPI to add `open?`  [dev docs](https://roamresearch.com/#/app/developer-documentation/page/A_vHEjJ7J)

Testable at: https://relemma-git-renderblockadditions.roamresearch.com/

https://www.loom.com/share/431502557a0e48cc88ab09d0c0cc7a43

### DIFF
--- a/src/components/Kanban.tsx
+++ b/src/components/Kanban.tsx
@@ -19,6 +19,7 @@ import toCellValue from "../utils/toCellValue";
 import extractTag from "roamjs-components/util/extractTag";
 import deleteBlock from "roamjs-components/writes/deleteBlock";
 import getSubTree from "roamjs-components/util/getSubTree";
+import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
 
 const zPriority = z.record(z.number().min(0).max(1));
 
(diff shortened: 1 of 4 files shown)


src/components/Kanban.tsx
```diff
@@ -98,13 +178,18 @@ const Kanban = ({
...
   resultKeys: Column[];
   data: Result[];
   layout: Record<string, string | string[]>;
   onQuery: () => void;
   parentUid: string;
+  views: { column: string; mode: string; value: string }[];
 }) => {
+  const [localFoldStates, setLocalFoldStates] = useState(
+    new Map<string, boolean>()
+  );
```
mdroidian, line 192: a lot of changes in this PR are from adding `localFoldStates`.

Problem this solves:
- user selects `embed` in `Kanban` view and chooses `embed` `open?` state
- user manually changes one or more block fold states
- user performs an action that causes a re-render
  - this includes hovering over the column menu or dragging and dropping cards
- manually changed blocks then revert to chosen `embed` state

This seems to happen less in `Table` view which leads me to believe this might be solved by investigating and addressing the root cause of the re-renders instead of adding `localFoldStates`, but I'm not certain.

src/components/ResultsView.tsx
```diff
@@ -395,134 +399,110 @@ const ResultsView: ResultsViewComponent = ({
...
-            }
-            onInteraction={(next, e) => {
-              if (!e) return;
-              const target = e.target as HTMLElement;
-              if (
-                target.classList.contains("bp3-menu-item") ||
-                target.closest(".bp3-menu-item")
-              ) {
-                return;
+          <div
```
mdroidian, line 412: This fixed a separate issue where any `ResultsView` mode would re-render multiple times when moving the mouse over the component.

mdroidian: video showing the initial render issue and the trial of `stopPropagation()` as a fix

https://www.loom.com/share/89aef4d1bb9742249b97041fd5a73fdf

src/components/Kanban.tsx
```diff
@@ -19,22 +19,97 @@ import toCellValue from "../utils/toCellValue";
...
+        className="roamjs-query-embed rounded-xl bg-white p-4 "
+        onClick={(e) => {
+          const target = e.target as HTMLElement;
+          if (!target.classList.contains("rm-caret")) return;
+
+          const kanbanCard = target.closest(".roamjs-kanban-card");
+          const uid = kanbanCard?.getAttribute("data-uid");
+          if (!uid) return;
+
+          const isOpen = target.classList.contains("rm-caret-open");
```
dvargas92495, line 89: see if there's a way to get this from `roamAlphaAPI`

src/components/Kanban.tsx
dvargas92495: Pull out `CellEmbed` into its own component and see if you could live without the added state

dvargas92495 (approved):

PR #185, opened 2023-12-03
Global exact match filters

mdroidian:
Currently exact match is only available on ResultsTable via the Roam filter component ( which has been deprecated in QB )

image

This would allow that functionality to all other views.

- [ ] handle trimmed edge case via `getSettingValuesFromTree()`


src/components/MultiSelect.tsx
```diff
@@ -0,0 +1,37 @@
...
+          text={props.transformItem ? props.transformItem(item) : item}
+          active={modifiers.active}
+          onClick={handleClick}
+        />
+      )}
+      tagRenderer={(item) =>
+        props.transformItem ? props.transformItem(item) : item
+      }
+      itemPredicate={props.filterable ? itemPredicate : undefined}
+    ></MultiItemSelect>
```
dvargas92495, line 33: .

src/components/ResultsView.tsx
```diff
@@ -170,7 +173,10 @@ const QueryUsed = ({ parentUid }: { parentUid: string }) => {
   );
 };
 
-const SUPPORTED_COLUMN_FILTER_TYPES = [{ id: "contains" }];
+const SUPPORTED_COLUMN_FILTER_TYPES = [
+  { id: "contains" },
+  { id: "contains exactly" },
```
dvargas92495, line 178: `equals`

PR #190, opened 2023-12-08
Ability to add source to canvas node

mdroidian:
This PR lets the user "link" a referenced to canvas node via
 - dialog create
 - dialog edit
 - dragging an arrow from source to target

https://www.loom.com/share/9034601384d349a4a73e0e83964d5a26

### DIFF
--- a/src/components/DiscourseNodeCanvasSettings.tsx
+++ b/src/components/DiscourseNodeCanvasSettings.tsx
@@ -13,18 +13,23 @@ import getBasicTreeByParentUid from "roamjs-components/queries/getBasicTreeByPar
 import getSettingValueFromTree from "roamjs-components/util/getSettingValueFromTree";
 import setInputSetting from "roamjs-components/util/setInputSetting";
 
+export const formatHexColor = (color: string) => {
+  if (!color) return "";
+  const COLOR_TEST = /^[0-9a-f]{6}$/i;
+  if (color.startsWith("#")) {
+    // handle legacy color format
+    return color;
+  } else if (COLOR_TEST.test(color)) {
+    return "#" + color;
+  }
+  return "";
+};
+
 const DiscourseNodeCanvasSettings = ({ uid }: { uid: string }) => {
   const tree = useMemo(() => getBasicTreeByParentUid(uid), [uid]);
   const [color, setColor] = useState<string>(() => {
     const color = getSettingValueFromTree({ tree, key: "color" });
-    const COLOR_TEST = /^[0-9a-f]{6}$/i;
-    if (color.startsWith("#")) {
-      return color;
-      // handle legacy color format
-    } else if (COLOR_TEST.test(color)) {
-      return "#" + color;
-    }
-    return "";
+    return formatHexColor(color);
   });
   const [alias, setAlias] = useState<string>(() =>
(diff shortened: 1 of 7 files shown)


src/components/DiscourseNodeCanvasSettings.tsx
```diff
@@ -13,18 +13,23 @@ import getBasicTreeByParentUid from "roamjs-components/queries/getBasicTreeByPar
...
     const color = getSettingValueFromTree({ tree, key: "color" });
-    const COLOR_TEST = /^[0-9a-f]{6}$/i;
-    if (color.startsWith("#")) {
-      return color;
-      // handle legacy color format
-    } else if (COLOR_TEST.test(color)) {
-      return "#" + color;
-    }
-    return "";
+    return formatHexColor(color);
```
mdroidian, line 32: bug fix

src/components/TldrawCanvas.tsx
```diff
@@ -554,51 +555,49 @@ class DiscourseNodeUtil extends TLBoxUtil<DiscourseNodeShape> {
...
               }
 
+              // Update Shape Props
+              setSizeAndImgProps({ context: this, text, uid });
+              this.updateProps(shape.id, {
+                title: text,
+                uid,
+              });
+
+              // Update Shape Relations
```
mdroidian, line 586: should just be a refactor

src/components/TldrawCanvas.tsx
```diff
@@ -944,107 +1079,196 @@ const TldrawCanvas = ({ title }: Props) => {
...
+                            sourceTitle.trim();
+                          if (!isTargetTitleCurrent || !isSourceTitleCurrent) {
+                            return deleteAndWarn(
+                              "Either the source or target node has been renamed. Please update the nodes and try again."
+                            );
+                          }
+
+                          // Hack for default shipped EVD format: [[EVD]] - {content} - {Source},
+                          // replace when migrating from format to specification
+                          let newTitle: string;
```
dvargas92495, line 1126:
```typescript
const format = "[[EVD]] - {content} - {Source}";
const newTitle = format.replace("{content}", targetTitle).replace("{Source}", `[[${sourceTitle}]]`)
```

src/components/TldrawCanvas.tsx
```diff
@@ -1401,6 +1636,12 @@ const TldrawCanvas = ({ title }: Props) => {
                     name,
                   ])
                 ),
+                ...Object.fromEntries(
+                  allAddReferencedNodeActions.map((name) => [
+                    `shape.relation.${name}`,
```
dvargas92495, line 1641: give this its own shape prefix

dvargas92495 (approved):

src/components/TldrawCanvas.tsx
mdroidian: Ah.  `targetTitle` wouldn't just be `{content}`, it would contain the `[[EVD]] - `, so the output would be `[[EVD]] - [[EVD]] - `

I'm not sure how to grab `{content}` based on the arbitrary format string.  I believe this needs to be solved with specification.

Maybe we could create a dynamic regex pattern to capture `{content}`? I imagine it would be similar to how format is converted to `has title` in specification currently.

PR #201, opened 2023-12-16
add activeSort support for Kanban view

mdroidian:
I'm unsure how we should handle dragging when `activeSort` is set.

Use cases:

Drag to new column to change attribute
 - my intuition says that `activeSort` should be maintained

Drag to reposition in same column
 - currently not supported, `activeSort` is maintained
 - [x] probably should remove `activeSort` and create a new prioritization based on where cards currently are + moved card.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "query-builder",
-  "version": "1.21.2",
+  "version": "1.22.0",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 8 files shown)


dvargas92495 (approved):

mdroidian:
https://www.loom.com/share/fc734cfe0d90418cb42f4bcf7d82f9e1

dvargas92495 (approved):

PR #204, opened 2023-12-21
Add "has uid" as condition

mdroidian:


dvargas92495:
why do we want this? at a high level I was hoping to abstract away uids from the user - they usually are trying to query _by some other property of a block_ and not by its id.

at a base level, I'd probably prefer `has block reference` or something like that

mdroidian:
> why do we want this? at a high level I was hoping to abstract away uids from the user - they usually are trying to query _by some other property of a block_ and not by its id.
> 
> at a base level, I'd probably prefer `has block reference` or something like that

I've reached for this a few times, but I forget the exact use cases for all of those.  
But one specifically is with a smartblock workflow of pasting text into the current block:
 - if current block has ancestor that references a link, append that link to the pasted text

PR #205, opened 2023-12-21
Optimize Discourse Export

mdroidian:
Just before creating this PR I realized I missed something 😑 (which is the reason for all the comment edits)

I thought I had an epiphany here (see video 😅) but in my dev branch I wasn't accounting for the results being passed by the `Export Discourse Graph` command.  But it turns out even with those results being passed in, because of `block` / `page`, the `gatherings` output gets significantly increased.  (even more than my video shows)

https://www.loom.com/share/664fd3cc2125492cbb16009500b9d0a8

Looks like the `Set` check clears it up.

And I'm curious, @dvargas92495, what is the purpose of the `matchDiscourseNode` filter in `getPageData`? (see comment in review)

### DIFF
--- a/src/utils/getExportTypes.ts
+++ b/src/utils/getExportTypes.ts
@@ -195,9 +195,16 @@ const toMarkdown = ({
 type Props = {
   results?: ExportDialogProps["results"];
   exportId: string;
+  isExportDiscourseGraph: boolean;
 };
 
-const getExportTypes = ({ results, exportId }: Props): ExportTypes => {
+export type DiscourseExportResult = Result & { type: string };
+
+const getExportTypes = ({
+  results,
+  exportId,
+  isExportDiscourseGraph,
+}: Props): ExportTypes => {
   const allRelations = getDiscourseRelations();
   const allNodes = getDiscourseNodes(allRelations);
   const nodeLabelByType = Object.fromEntries(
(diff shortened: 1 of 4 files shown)


src/utils/getExportTypes.ts
```diff
@@ -235,7 +236,13 @@ const getExportTypes = ({ results, exportId }: Props): ExportTypes => {
...
           }))
       )
-        .filter(({ text }) => matchDiscourseNode({ title: text, ...n }))
+        .filter(({ text }) => {
+          if (matchedTexts.has(text)) return false;
+          const isMatch = matchDiscourseNode({ title: text, ...n });
+          if (isMatch) matchedTexts.add(text);
+
+          return isMatch;
+        })
```
mdroidian, line 245: What is the purpos of the original matchDiscourseNode filter?
I'm likely missing something, because for the `Export Discourse Graph` command, it seems to just be double up the work?

mdroidian: Ah, to get the type maybe?

It seems like an expensive loop (`allNodes` mapped over all results running `matchDiscourseNode` which hits `roamAlphaAPI`) when we I believe we have the type at initialization of `Export Discourse Graph`.

mdroidian: I made some changes to pass the `type` with Export Discourse Graph `results`: https://github.com/RoamJS/query-builder/pull/205/commits/34f2311fcc9d005587762e0a78ade17b8d536efc so we can bypass this loop altogether.

mdroidian: https://www.loom.com/share/ddd3946b3db94545b89ffa73bfa14aee

dvargas92495: yea I think is to get the discourse type - would definitely prefer bypassing if we could

dvargas92495 (approved):
Have I showed how to use performance monitor? Would be a more efficient way of seeing problematic functions over creating custom UIs haha

PR #206, opened 2023-12-23
bug fix: add check for shape.type on dragging_handle

mdroidian:
bug introduced here: https://github.com/RoamJS/query-builder/pull/190

couldn't add a regular arrow that wasn't connected to non-shapes and lines console errored

https://www.loom.com/share/bb176761e62d4c1ab765670378a37a71

### DIFF
--- a/src/components/TldrawCanvas.tsx
+++ b/src/components/TldrawCanvas.tsx
@@ -1047,6 +1047,10 @@ const TldrawCanvas = ({ title }: Props) => {
                 return TLSelectTool.children().map((c) => {
                   if (c.id === "dragging_handle") {
                     const Handle = c as unknown as typeof DraggingHandle;
+                    const allRelationIdSet = new Set(allRelationIds);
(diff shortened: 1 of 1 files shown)


src/components/TldrawCanvas.tsx
```diff
@@ -1047,15 +1047,27 @@ const TldrawCanvas = ({ title }: Props) => {
...
                           type: "misc",
                           name: "complete",
                         });
 
-                        const arrow = this.app.getShapeById(this.shapeId);
-                        if (!arrow) return;
+                        const shape = this.app.getShapeById(this.shapeId);
+                        if (!shape) return;
+                        // TODO: find a better way to identify custom arrow shapes
+                        // shape.type or shape.name probably?
```
dvargas92495, line 1064: could use `roamAlphaAPI.pull` but we already have the set in scope so might as well use it as you have it

dvargas92495 (approved):

PR #218, opened 2024-01-05
Tldraw - optimize placement on node import 

mdroidian:
https://www.loom.com/share/3c109f4a076c4143acaa2098b2554038

### DIFF
--- a/src/components/Export.tsx
+++ b/src/components/Export.tsx
@@ -37,7 +37,7 @@ import getRoamUrl from "roamjs-components/dom/getRoamUrl";
 import findDiscourseNode from "../utils/findDiscourseNode";
 import { DEFAULT_CANVAS_PAGE_FORMAT } from "../index";
 import { createShapeId } from "@tldraw/tlschema";
-import { defaultDiscourseNodeShapeProps } from "./TldrawCanvas";
+import { MAX_WIDTH } from "./TldrawCanvas";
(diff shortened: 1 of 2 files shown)


src/components/Export.tsx
```diff
@@ -170,10 +176,65 @@ const ExportDialog: ExportDialogComponent = ({
...
+    };
+    const shapeBounds = extractShapesBounds(tldraw);
+
+    type CommonBounds = {
+      top: number;
+      right: number;
+      bottom: number;
+      left: number;
+    };
+    const findCommonBounds = (shapes: ShapeBounds[]): CommonBounds => {
```
dvargas92495, line 211: surprised tldraw doesnt expose something for this

dvargas92495 (approved):

src/components/Export.tsx
mdroidian: They do, `app.allShapesCommonBounds` but I don't know how to get the `app` (the tlddraw instance of a canvas that isn't loaded) in `Export.tsx`.  

Did a little digging and found `useEditor`, but I believe you can only use that in children of the tldraw component.

dvargas92495: oh damn. is this method just a copy paste of what they're doing in there?

mdroidian: Nah, I figured it was easy enough to make so just did it.

PR #219, opened 2024-01-05
Filter out default discourse nodes on dgraph init (Closes #198)

### LINKED ISSUE #198: Fresh install of Discourse Graph bugs
issues
1. does not add any grammar
2. adds 3 relations with no `source` or `destination`
3. if there is a relation that is has an undefined `source`/`destination`, `Discourse Context` will fail

Steps to repo for discourse context failing
1. create new graph
2. install query builder
3. enable discourse graph
4. go to `[[roam/js/discourse-graph]]` (for issues 1 and 2)
5. create two grammar nodes
6. update a relation with the new grammar nodes as `source` and `destination`
7. create a relation in any page
8. view node's `discourse context`

TODO
- [x] fix default grammar
- [x] fix default relations

mdroidian:


dvargas92495 (approved):
oh snap great catch

PR #220, opened 2024-01-06
Add focus to input element when adding referenced node

mdroidian:
https://www.loom.com/share/26749dd0c3ae49e69a72ae0cc921b770

### DIFF
--- a/src/components/TldrawCanvasLabelDialog.tsx
+++ b/src/components/TldrawCanvasLabelDialog.tsx
@@ -107,6 +107,14 @@ const LabelDialogAutocomplete = ({
     setOptions,
     setReferencedNodeOptions,
   ]);
+  useEffect(() => {
+    if (isAddReferencedNode) {
+      const inputElement = document.getElementById(
+        "referenced-node-autocomplete-input"
+      );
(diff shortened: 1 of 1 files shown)


src/components/TldrawCanvasLabelDialog.tsx
```diff
@@ -254,6 +262,7 @@ const LabelDialogAutocomplete = ({
           <div className="referenced-node-autocomplete">
             <Label>{referencedNode?.name}</Label>
             <AutocompleteInput
+              id="referenced-node-autocomplete-input"
```
dvargas92495, line 265: [nit] this could break with multiple canvas elements on the page

the "react"-y way is usually:
- create a `cont inputRef = useRef()`
- pass the ref to AutocompleteInput `ref={inputRef}`
- call `inputRef.current.focus()` in the `useEffect`

Now this is prohibited by `AutocompleteInput` not supporting forwarding refs (which we should eventually add, low priority and would be tricky). The workaround to that could be to pass the ref into the containing `div` and then do `div.getElementsByTagName("input")[0].focus()`

dvargas92495 (approved):

PR #223, opened 2024-01-11
Add support for greater than and less than column filter types

mdroidian:
Tests (*no sound)
https://www.loom.com/share/c355a731a3ac4228b946eeecadbd1254

### DIFF
--- a/src/utils/postProcessResults.ts
+++ b/src/utils/postProcessResults.ts
@@ -3,19 +3,18 @@ import { Result } from "roamjs-components/types/query-builder";
 import extractTag from "roamjs-components/util/extractTag";
 import parseResultSettings from "./parseResultSettings";
 
+const transform = (_val: Result[string]) =>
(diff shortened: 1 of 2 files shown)


src/utils/postProcessResults.ts
```diff
@@ -83,6 +82,14 @@ const postProcessResults = (
...
             }
             return columnFilter.value.some((v) => r[columnFilter.key] === v);
+          case "greater than":
+          case "less than":
+            const transformedFilter = transform(columnFilter.value[0]);
+            const transformedResult = transform(r[columnFilter.key]);
+
+            return columnFilter.type === "greater than"
+              ? transformedResult > transformedFilter
+              : transformedResult < transformedFilter;
```
dvargas92495, line 92: we are already checking the `columnFilter.type === "greater than"` case, so I typically keep each "function" in each case

dvargas92495 (approved):

PR #225, opened 2024-01-15
handle custom first column name

mdroidian:
when the first column is renamed via a `node` selection, `r.text` was undefined in `Share Data` > `Send To`
image

image

### DIFF
--- a/src/components/Export.tsx
+++ b/src/components/Export.tsx
@@ -245,7 +245,7 @@ const ExportDialog: ExportDialogComponent = ({
       const nodeType = discourseNode ? discourseNode.type : "page-node";
       const extensionAPI = getExtensionAPI();
       const { h, w, imageUrl } = await calcCanvasNodeSizeAndImg({
-        text: r[firstColumnKey],
+        nodeText: String(r[firstColumnKey]),
         uid: r.uid,
         nodeType,
         extensionAPI,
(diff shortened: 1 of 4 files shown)


src/components/Export.tsx
```diff
@@ -316,21 +321,48 @@ const ExportDialog: ExportDialogComponent = ({
...
+      else addToSelectedPage();
+      renderToast({
+        content: "Results sent!",
+        intent: "success",
+        id: "query-builder-export-success",
+      });
+    } catch (e) {
+      const error = e as Error;
+      renderToast({
+        content: "Looks like there was an error.  The team has been notified.",
```
dvargas92495, line 343: [nit] looks like an unintentional extra space here

src/utils/calcCanvasNodeSizeAndImg.ts
```diff
@@ -44,7 +44,7 @@ const calcCanvasNodeSizeAndImg = async ({
   nodeType,
   extensionAPI,
 }: {
-  text: string;
+  text: string | number | Date;
```
dvargas92495, line 47: [nit] `text` has a distinct meaning throughout the codebase(s) as the string value of a roam block. I'd rather we:
- cast to string before inserting into this method
- calling it `value` or something like that

dvargas92495 (approved):

src/components/Export.tsx
mdroidian: hah!  I'm still in the 'two spaces after a period' era apparently. Muscle memory can be a bitch.

PR #227, opened 2024-01-18
Kanban - Change max columns and update on group by/display change

mdroidian:
https://www.loom.com/share/a7468e711ac74aee95533f85a704e773

### DIFF
--- a/src/components/ResultsView.tsx
+++ b/src/components/ResultsView.tsx
@@ -597,6 +597,7 @@ const ResultsView: ResultsViewComponent = ({
                               value,
                               blockUid: layoutNode.uid,
                             });
+                            // TODO: replace this with an `onSettingChange` in SUPPORTED_LAYOUTS
                             if (s.options === "columns") {
                               setInputSettings({
                                 key: "columns",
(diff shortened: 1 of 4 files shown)


src/components/ResultsView.tsx
```diff
@@ -597,6 +597,13 @@ const ResultsView: ResultsViewComponent = ({
                               value,
                               blockUid: layoutNode.uid,
                             });
+                            if (s.options === "columns") {
+                              setInputSettings({
+                                key: "columns",
+                                blockUid: layoutNode.uid,
+                                values: [],
+                              });
+                            }
```
mdroidian, line 606: Not *super* happy with hardcoding `"columns"` in here again ... if you have any suggestions, I'm all ears.

mdroidian: This is for clearing out a layout setting when certain layout specific settings are updated.
specifically these two for Kanban
```
      { key: "key", label: "Group By", options: "columns" },
      { key: "display", label: "Display", options: "columns" },
```
RE: subtree
I wonder how ugly the backwards compatibility would be for this with multiple settings already

RE: `clearOnSwitch` 
I like this, I think this would work

PR #236, opened 2024-01-30
Add PDF Export type

mdroidian:
requires https://github.com/samepage-network/samepage.network/pull/126

### DIFF
--- a/src/components/Export.tsx
+++ b/src/components/Export.tsx
@@ -386,7 +386,7 @@ const ExportDialog: ExportDialogComponent = ({
     try {
       const response = await apiPost({
         domain,
-        path: "pdf/query-builder",
+        path: "pdf",
         data: {
           files: preparedFiles,
           filename,
--- a/src/components/Export.tsx
+++ b/src/components/Export.tsx
@@ -496,15 +496,59 @@ const ExportDialog: ExportDialogComponent = ({
                     if (!files.length) {
                       setDialogOpen(true);
                       setError("Failed to find any results to export.");
-                    } else {
-                      files.forEach(({ title, content }) =>
-                        zip.file(title, content)
-                      );
-                      zip.generateAsync({ type: "blob" }).then((content) => {
-                        saveAs(content, `${filename}.zip`);
-                        onClose();
-                      });
+                      return;
                     }
+
+                    if (activeExportType === "PDF") {
+                      const response = await apiPost({
+                        domain: "http://localhost:3003",
+                        path: "pdf",
+                        data: {
+                          files,
+                          method: "michael-error",
+                          type: "michael",
+                          version: process.env.VERSION,
+                          notebookUuid: JSON.stringify({
+                            owner: "RoamJS",
+                            app: "query-builder",
+                            workspace: window.roamAlphaAPI.graph.name,
+                          }),
+                        },
+                      }).catch((error) => {
+                        console.error("Error fetching ZIP from backend", error);
+                        setError("Failed to export files.");
+                      });
+
+                      if (response) {
+                        const parsedResponse = JSON.parse(response.data);
+                        const base64ToBlob = (
+                          base64: string,
+                          type = "application/octet-stream"
+                        ) => {
+                          const binaryString = window.atob(base64);
+                          const len = binaryString.length;
+                          const bytes = new Uint8Array(len);
+                          for (let i = 0; i < len; i++) {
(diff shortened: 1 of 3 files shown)


src/components/Export.tsx
```diff
@@ -365,6 +366,49 @@ const ExportDialog: ExportDialogComponent = ({
...
+        const len = binaryString.length;
+        const bytes = new Uint8Array(len);
+        for (let i = 0; i < len; i++) {
+          bytes[i] = binaryString.charCodeAt(i);
+        }
+        return new Blob([bytes], { type: type });
+      };
+      const blob = base64ToBlob(parsedResponse.body, "application/zip");
+      saveAs(blob, `${filename}.zip`);
+      onClose();
```
dvargas92495, line 409: this conversion looks expensive - here's a possible alternative:
- PDF conversion is done and uploaded to a public link `.zip` in samepage's backend
- we download straight from that link here

src/utils/getExportTypes.ts
```diff
@@ -88,7 +90,7 @@ const getFilename = ({
     removeSpecialCharacters
       ? baseName.replace(/[<>:"/\\|\?*[\]]/g, "")
       : baseName
-  }.md`;
+  }.${extension}`;
```
dvargas92495, line 93: looks like the default is `..md`

src/utils/getExportTypes.ts
```diff
@@ -596,6 +642,162 @@ const getExportTypes = ({
         ];
       },
     },
+    {
+      name: "PDF",
```
dvargas92495, line 646: it looks like this method is just doing the markdown export and then converting to pdf - is that right? could we reuse?

dvargas92495 (approved):

src/utils/getExportTypes.ts
mdroidian: Much of it, yes, but not all.  PDF doesn't use the YAML functionality.

I actually started to move these functions out (as I did with `getExportSettings()`):
-  `treeNodeToMarkdown()`
- `getMarkdownContent()`
- `getDiscourseResultsContent()`
- `getReferenceResultsContent()`

but with the amount of rewrite to the markdown callback, it started to feel like it deserved it's own PR.

Should I include that with this PR or would adding a `// TODO` suffice?

dvargas92495: discussed offline, good to defer

src/components/Export.tsx
mdroidian: Do we already have a utility for downloading files from samepage's backend (s3) from a frontend extension? (I didn't see one)

Or do we need to create a lambda to generate a signed url?  Or something else?

dvargas92495: The expectation is that we don't need a lambda - as long as the path matches one of these origin paths that it should download: https://github.com/samepage-network/samepage.network/blob/main/app/server/index.ts

mdroidian: Sorry, I don't follow.

Currently I'm using `uploadFileContent` to upload to s3 @ `data/pdfs`.
I'm not sure how to use `app/server/index` to retrieve a file uploaded to there.

Would I need to add `data/pdfs` to the `originPaths` in `app/server/index`?
Then create something like `/api/pdf/get.ts` to invoke `app/server/index`'s handler?

mdroidian: I tried adding `/^\/data\/pdfs\/.+$/,` to `originPaths` then retrieving in query builder with:

```typescript
apiGet<ArrayBuffer>({
        domain: `https://api.samepage.network`,
        path: `data/pdfs/${file}`
        buffer: true,
      });
```
      
But this results in a CORS error.
      
I have a feeling that `https://api.samepage.network` is incorrect
Is there a separate url for retrieving from s3?

dvargas92495: ah it would just be from `https://samepage.network` - the `originPaths` within `app/server/index` define all the paths where we simply just redirect the request to retrieve the content from S3.

As a general rule, we should try to avoid putting any "public assets" within the `data` directory, and instead, put the pdfs in likely a new top level `pdfs`

mdroidian: Excellent.

After adding `/^\/pdfs\/.*/`, `localhost` kept giving me `CORS` errors.  So I used `/videos` to test.
Just pinging you on this because there are some test uploads in that route, but I don't believe I have access to delete them.

PR #240, opened 2024-03-26
get current roam node content when dialog opens

mdroidian:
https://www.loom.com/share/cf6b07dcce304f59b5a7425ce383724b

### DIFF
--- a/src/components/TldrawCanvasLabelDialog.tsx
+++ b/src/components/TldrawCanvasLabelDialog.tsx
@@ -287,7 +287,7 @@ const LabelDialogAutocomplete = ({
 };
 
 type NodeDialogProps = {
-  label: string;
+  isExistingCanvasNode: boolean;
   onSuccess: (a: Result) => Promise<void>;
   onCancel: () => void;
   nodeType: string;
@@ -295,10 +295,14 @@ type NodeDialogProps = {
   discourseContext: DiscourseContextType;
 };
 
+const getCurrentNodeContent = (uid: string) => {
+  return getPageTitleByPageUid(uid) || getTextByBlockUid(uid);
+};
+
 const LabelDialog = ({
   isOpen,
   onClose,
-  label: _label,
+  isExistingCanvasNode,
   onSuccess,
   onCancel,
   nodeType,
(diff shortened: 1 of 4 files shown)


src/components/TldrawCanvasLabelDialog.tsx
```diff
@@ -306,15 +308,19 @@ const LabelDialog = ({
   const containerRef = useRef<HTMLDivElement>(null);
   const [error, setError] = useState("");
   const initialLabel = useMemo(() => {
-    if (_label) return _label;
+    if (_label)
+      return getPageTitleByPageUid(initialUid) || getTextByBlockUid(initialUid);
```
dvargas92495, line 312: why does truthiness of `_label` dictate to use data from initialUid instead?

src/components/TldrawCanvasLabelDialog.tsx
```diff
@@ -306,15 +308,19 @@ const LabelDialog = ({
...
     return getPlainTitleFromSpecification({ specification, text });
-  }, [_label, nodeType]);
+  }, [_label, nodeType, initialUid, isOpen]);
   const initialValue = useMemo(() => {
     return { text: initialLabel, uid: initialUid };
   }, [initialLabel, initialUid]);
-  const [label, setLabel] = useState(initialValue.text);
+  const [label, setLabel] = useState("");
+  useEffect(() => {
+    setLabel(initialLabel);
```
dvargas92495, line 322: I think bc `_label` updates, `initialLabel` updates, so doesn't this fire on every keystroke?

src/components/TldrawCanvasLabelDialog.tsx
mdroidian: it's a proxy for "Is this a new canvas node or an existing canvas node?"

If it is existing, get most up to date node content.
If it's new, get the `specification` text instead

It's not used elsewhere so I renamed it to `isExistingCanvasNode`

src/components/TldrawCanvasLabelDialog.tsx
mdroidian: I think `_label` only updated when `shape.props.title` changed, which is `onSuccess` of the dialog.

https://www.loom.com/share/09f71b10c61b4dbaa8aad2bbb61f0308

Regardless, testing this showed that it was firing for every node on canvas load.  Changed it to fire on `isOpen`

dvargas92495 (approved):

PR #242, opened 2024-03-28
add isDiscourseNode to dev api

mdroidian:


dvargas92495 (approved):
Who/what's planning to use this?

mdroidian:
> Who/what's planning to use this?

https://github.com/sid597/roam-llm-chat
An LLM Chat plugin that grabs query builder results, their content, and the content of their linked references, then sends that as context.  

The goal of this PR is to give the ability to limit the context by only sending linked references that are discourse nodes.

It'll probably be added here:
https://github.com/sid597/roam-llm-chat/blob/22363b371a8f3acb828415275a727ff052e509fa/src/main/ui/extract_data/chat.cljs#L169-L191

PR #243, opened 2024-04-02
add fireQuerySync to window object

mdroidian:
PR'd to build `/release` for client testing.


dvargas92495 (approved):
whatever the use case is, I bet they'll make it async eventually 😄

PR #246, opened 2024-04-06
Add referenced node to exported discourse context

mdroidian:
From a client request to add referenced nodes to exports.

EG: Format = [[EVD]] - {content} - {Source}
then add "Source: [[The source page title]]" to the appended Discourse Context on export

image

Another workaround for `format` over `specification`, re: https://github.com/RoamJS/query-builder/issues/189 😞

### DIFF
--- a/src/utils/formatUtils.ts
+++ b/src/utils/formatUtils.ts
@@ -8,6 +8,10 @@ import discourseNodeFormatToDatalog from "./discourseNodeFormatToDatalog";
 import createOverlayRender from "roamjs-components/util/createOverlayRender";
 import { render as renderToast } from "roamjs-components/components/Toast";
 import FormDialog from "roamjs-components/components/FormDialog";
+import { QBClause, Result } from "./types";
+import findDiscourseNode from "./findDiscourseNode";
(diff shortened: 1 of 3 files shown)


src/utils/formatUtils.ts
```diff
@@ -109,3 +122,41 @@ export const getReferencedNodeInFormat = ({
...
+  const specification = discourseNode.specification;
+  const titleCondition = specification.find(
+    (s): s is QBClause => s.type === "clause" && s.relation === "has title"
+  );
+  if (!titleCondition) return null;
+
+  // Remove leading and trailing slashes and start/end modifiers
+  const patternStr = titleCondition.target.slice(1, -1).replace(/^\^|\$$/g, "");
+
+  // Since we aummar there's always a (.*?), we replace it with a specific pattern to capture text within [[ ]]
```
dvargas92495, line 148: > aummar

dvargas92495 (approved):

src/utils/formatUtils.ts
mdroidian: I don't always `aummar`, but when I do, I prefer to leave it in a comment. 😅

PR #248, opened 2024-04-17
add Github destination

mdroidian:
- [ ] requires https://github.com/samepage-network/samepage.network/pull/141

https://www.loom.com/share/0a446807534446558d9b160969a76e81

### DIFF
--- a/src/components/ExportGithub.tsx
+++ b/src/components/ExportGithub.tsx
@@ -1,7 +1,8 @@
 import { Button } from "@blueprintjs/core";
-import React, { useEffect, useRef, useState } from "react";
+import React, { useCallback, useEffect, useRef, useState } from "react";
 import MenuItemSelect from "roamjs-components/components/MenuItemSelect";
 import apiGet from "roamjs-components/util/apiGet";
+import localStorageGet from "roamjs-components/util/localStorageGet";
 import localStorageSet from "roamjs-components/util/localStorageSet";
 
 type UserReposResponse = {
@@ -26,7 +27,6 @@ export const ExportGithub = ({
   selectedRepo,
   setSelectedRepo,
   setError,
-  clearGitHubToken,
   gitHubAccessToken,
   setGitHubAccessToken,
   setCanSendToGitHub,
@@ -35,7 +35,6 @@ export const ExportGithub = ({
   selectedRepo: string;
   setSelectedRepo: (selectedRepo: string) => void;
   setError: (error: string) => void;
-  clearGitHubToken: boolean;
   gitHubAccessToken: string | null;
   setGitHubAccessToken: (gitHubAccessToken: string | null) => void;
   setCanSendToGitHub: (canSendToGitHub: boolean) => void;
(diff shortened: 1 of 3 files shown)


src/components/ExportGithub.tsx
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }, [gitHubAccessToken, isGitHubAppInstalled, selectedRepo]);
+
+  // clear github token if error writing to github
+  useEffect(() => {
+    if (clearGitHubToken) clearAuth();
+  }, [clearGitHubToken]);
+
+  const InstallButton = () => (
+    <Button
```
dvargas92495, line 146: [nit] We lose some benefits of React's memoization when we define these as new functions each time instead of having them as inline components below. could be further memoized with useCallbacks for each of the onClicks

src/components/ExportGithub.tsx
```diff
@@ -0,0 +1,211 @@
...
+        clearAuth();
+      }
+      return false;
+    }
+  };
+
+  // listen for messages from the auth window
+  useEffect(() => {
+    const handleGitHubAuthMessage = (event: MessageEvent) => {
+      if (event.data) {
```
dvargas92495, line 88: [nit] I believe we should be checking the origin this came from, otherwise malicious sites could post messages here. Check out `ExternalLogin`

src/components/ExportGithub.tsx
```diff
@@ -0,0 +1,211 @@
...
+      const res = await apiGet<{ installations: { app_id: number }[] }>({
+        domain: "https://api.github.com",
+        path: "user/installations",
+        headers: {
+          Authorization: `token ${gitHubAccessToken}`,
+        },
+      });
+      const installations = res.installations;
+      const isAppInstalled = installations.some(
+        (installation) => installation.app_id === 874256
```
dvargas92495, line 72: [nit] Let's set this value to an environment variable and include it in main.yaml

src/components/ExportGithub.tsx
```diff
@@ -0,0 +1,211 @@
...
+    localStorageSet("oauth-github", "");
+  };
+
+  const fetchAndSetInstallation = async () => {
+    try {
+      const res = await apiGet<{ installations: { app_id: number }[] }>({
+        domain: "https://api.github.com",
+        path: "user/installations",
+        headers: {
+          Authorization: `token ${gitHubAccessToken}`,
```
dvargas92495, line 67: [nit] I would access `gitHubAccessToken` from local storage which allows us to wrap this function as a useCallback

src/components/ExportGithub.tsx
```diff
@@ -0,0 +1,211 @@
...
+      });
+      const installations = res.installations;
+      const isAppInstalled = installations.some(
+        (installation) => installation.app_id === 874256
+      );
+      setIsGitHubAppInstalled(isAppInstalled);
+      return isAppInstalled;
+    } catch (error) {
+      const e = error as Error;
+      if (e.message === "Bad credentials") {
```
dvargas92495, line 78: [nit] is there an error code or a status code we could rely on instead?

dvargas92495 (approved):
🔥

src/components/ExportGithub.tsx
mdroidian: @dvargas92495 just want to confirm that the SamePage GitHub App ID can be stored in main.yaml as a number rather than a secret.

src/components/ExportGithub.tsx
mdroidian: This is what we are getting back from `apiGet`

```
{
    "message": "Bad credentials",
    "documentation_url": "https://docs.github.com/rest"
}
```

dvargas92495: ah damn - something we should improve in apiGet, feel free to defer

src/components/ExportGithub.tsx
mdroidian: resolved

PR #250, opened 2024-04-21
OR Branches = conditions fix and misc UX changes

mdroidian:
https://www.loom.com/share/5475bfd8d2d3495e8f0eae670a0f42b8

### DIFF
--- a/src/components/QueryEditor.tsx
+++ b/src/components/QueryEditor.tsx
@@ -182,16 +182,14 @@ const QueryNestedData = ({
 }) => {
   return (
     <>
-      <Button
-        rightIcon={"arrow-right"}
-        text={"Edit"}
-        onClick={() => setView({ uid: con.uid, branch: 0 })}
-        style={{
-          minWidth: 144,
-          maxWidth: 144,
-          paddingRight: 8,
-        }}
-      />
+      <span style={{ minWidth: 144, display: "inline-block" }}>
+        <Button
+          rightIcon={"arrow-right"}
+          text={"Edit"}
+          onClick={() => setView({ uid: con.uid, branch: 0 })}
+          style={{ maxHeight: 32 }}
+        />
+      </span>
       <span
         style={{
           minWidth: 144,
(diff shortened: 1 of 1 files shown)


src/components/QueryEditor.tsx
```diff
@@ -598,6 +596,123 @@ const QueryEditor: QueryEditorComponent = ({
...
+            uid,
+            source: "node",
+            relation: "",
+            target: "",
+            type: "clause",
+          },
+        ]);
+      } else {
+        setConditions([
+          ...conditions,
```
dvargas92495, line 679: why can't we use the argument version of the state setter here?

src/components/QueryEditor.tsx
```diff
@@ -598,6 +596,123 @@ const QueryEditor: QueryEditorComponent = ({
...
+    if (view.uid === parentUid) return;
+
+    const branches = viewCondition?.conditions.length;
+    const branchUid = getNthChildUidByBlockUid({
+      blockUid: view.uid,
+      order: view.branch,
+    });
+    const branchConditions = getChildrenLengthByPageUid(branchUid);
+
+    if (!branches) createBranch();
```
dvargas92495, line 610: [nit] we should be memoizing methods that enter a `useEffect` with `useCallback`s and making sure they are in the dependency array. If we are worried about `createBranch` updating too much, we could set it to a ref and access the ref inner value from the use effect

dvargas92495: so both this and `addCondition`

src/components/QueryEditor.tsx
```diff
@@ -598,6 +596,123 @@ const QueryEditor: QueryEditorComponent = ({
...
+          viewStack
+            .slice(0, -1)
+            .concat([{ uid: view.uid, branch: view.branch + 1 }])
+        );
+        event.preventDefault();
+      }
+    };
+    document.addEventListener("keydown", handleKeyDown);
+    return () => document.removeEventListener("keydown", handleKeyDown);
+  }, [view, viewCondition]);
```
dvargas92495, line 651: [oos] In general, we should look for a way to look up component state without having them as dependencies to use effect that are truly meant to be set up once (on mount). One strategy for this is the ref approach above.

dvargas92495 (approved):

PR #251, opened 2024-04-22
hardcode github app id

mdroidian:
`process` returning undefined.

Build process not catching it?  unsure.  Will inspect later.

### DIFF
--- a/src/components/ExportGithub.tsx
+++ b/src/components/ExportGithub.tsx
@@ -62,7 +62,7 @@ export const ExportGithub = ({
         },
       });
       const installations = res.installations;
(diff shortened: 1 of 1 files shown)


src/components/ExportGithub.tsx
```diff
@@ -62,7 +62,7 @@ export const ExportGithub = ({
         },
       });
       const installations = res.installations;
-      const APP_ID = Number(process.env.GITHUB_APP_ID);
+      const APP_ID = 312167; // TODO - pull from process.env.GITHUB_APP_ID
```
dvargas92495, line 65: `GITHUB_APP_ID` is likely missing from `main.yaml`

PR #252, opened 2024-05-07
Check samepage for access-token on Authorize

mdroidian:
This will be used when the extension doesn't allow for a popup window to pass a `postMessage` (eg: electron apps).

Requires: https://github.com/samepage-network/samepage.network/pull/154

### DIFF
--- a/src/components/ExportGithub.tsx
+++ b/src/components/ExportGithub.tsx
@@ -209,7 +209,7 @@ export const ExportGithub = ({
 
             let attemptCount = 0;
             const check = () => {
-              if (attemptCount < 10) {
+              if (attemptCount < 30) {
                 apiPost({
                   path: "access-token",
                   domain: isDev
--- a/src/components/ExportGithub.tsx
+++ b/src/components/ExportGithub.tsx
@@ -1,6 +1,12 @@
 import { Button } from "@blueprintjs/core";
 import nanoid from "nanoid";
-import React, { useCallback, useEffect, useRef, useState } from "react";
+import React, {
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+  useMemo,
+} from "react";
 import MenuItemSelect from "roamjs-components/components/MenuItemSelect";
 import apiGet from "roamjs-components/util/apiGet";
 import apiPost from "roamjs-components/util/apiPost";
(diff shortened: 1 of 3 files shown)


src/components/ExportGithub.tsx
```diff
@@ -44,14 +47,24 @@ export const ExportGithub = ({
   const [isGitHubAppInstalled, setIsGitHubAppInstalled] = useState(false);
   const [clickedInstall, setClickedInstall] = useState(false);
   const [repos, setRepos] = useState<UserRepos>(initialRepos);
+  const [state, setState] = useState("");
   const showGitHubLogin = isGitHubAppInstalled && !gitHubAccessToken;
   const repoSelectEnabled = isGitHubAppInstalled && gitHubAccessToken;
 
+  const isDev = getNodeEnv() === "development";
```
dvargas92495, line 54: [nit] `useMemo`

src/components/ExportGithub.tsx
```diff
@@ -177,11 +192,37 @@ export const ExportGithub = ({
...
             authWindow.current = window.open(
-              "https://github.com/login/oauth/authorize?client_id=Iv1.e7e282a385b7b2da",
+              `https://github.com/login/oauth/authorize?${params}`,
               "_blank",
               `width=${WINDOW_WIDTH}, height=${WINDOW_HEIGHT}, top=${WINDOW_TOP}, left=${WINDOW_LEFT}`
             );
+
+            let attemptCount = 0;
+            const check = () => {
+              if (attemptCount < 10) {
```
dvargas92495, line 206: let's do two min

dvargas92495 (approved):

src/components/ExportGithub.tsx
mdroidian: Two minutes seems excessive.  To be clear, this would mean the user would wait 2 minutes staring at "Please Wait" (if they didn't manually cancel) and we would send 120 `POST` requests.

What scenario would waiting two minutes be useful for?

dvargas92495: 30 seconds then? just trying to mitigate any scenario with spotty internet/slow devices, 10 seconds felt quick

mdroidian: Sold!

PR #253, opened 2024-05-09
New Feature: GitHub Sync

mdroidian:
https://www.loom.com/share/0f3a0e72bc424d2f99783ab33a73f5fc


src/components/GitHubSync.tsx
```diff
@@ -0,0 +1,947 @@
...
+                  description: "Select a node to pull comments from",
+                  options: {
+                    items: [
+                      "None",
+                      ...getDiscourseNodes()
+                        .map((node) => node.text)
+                        .filter((text) => text !== "Block"),
+                    ],
+                  },
+                  defaultValue: "None",
```
mdroidian, line 921: maybe this should create be a CustomPanel and create a block on select/enter so the user can have multiple defined?

src/components/GitHubSync.tsx
```diff
@@ -0,0 +1,947 @@
...
+            containerDiv
+          );
+        }
+      },
+    });
+
+    return () => {
+      headerCommentObserver.forEach((o) => o.disconnect());
+      childCommentObserver.forEach((o) => o.disconnect());
+    };
```
mdroidian, line 532: adding a console.log here didn't fire.  not sure why or if that means these observers aren't being cleaned up.

dvargas92495:
This is so cool!! I wanna give this one a thoughtful review, so making a note to revisit on Saturday

src/components/Export.tsx
```diff
@@ -211,6 +227,70 @@ const ExportDialog: ExportDialogComponent = ({
...
+      // https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#create-an-issue
+      const response = await apiPost({
+        domain: "https://api.github.com",
+        path: `repos/${selectedRepo}/issues`,
+        headers: {
+          Authorization: `token ${gitHubAccessToken}`,
+        },
+        data: {
+          title,
+          body,
```
dvargas92495, line 249: [oos] There will be formatting differences, like Roam bolding -> Github bolding.

This is where proxying through SamePage I think could be helpful, since we are defining the AtJson standard there to be interoperable across all text editors

src/components/ExportGithub.tsx
```diff
@@ -23,13 +17,41 @@ type UserReposResponse = {
...
+    // https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-app-installations-accessible-to-the-user-access-token
+    const res = await apiGet<{ installations: { app_id: number }[] }>({
+      domain: "https://api.github.com",
+      path: "user/installations",
+      headers: {
+        Authorization: `token ${localStorageGet("oauth-github")}`,
+      },
+    });
+    const installations = res.installations;
+    const APP_ID = isDev ? 882491 : 312167; // TODO - pull from process.env.GITHUB_APP_ID
```
dvargas92495, line 45: [oos] yes please

src/components/Export.tsx
```diff
@@ -501,6 +581,8 @@ const ExportDialog: ExportDialogComponent = ({
               gitHubAccessToken={gitHubAccessToken}
               setGitHubAccessToken={setGitHubAccessToken}
               setCanSendToGitHub={setCanSendToGitHub}
+              githubDestination={githubDestination}
+              setGithubDestination={setGithubDestination}
```
dvargas92495, line 585: [nit] in general, use case specific state/props and callbacks should be defined within use case specific components. is there a way we could push all of this state within the component instead?

src/components/ExportGithub.tsx
```diff
@@ -23,13 +17,41 @@ type UserReposResponse = {
   ];
   status: number;
 };
-type UserRepos = UserReposResponse["data"];
-const initialRepos: UserRepos = [{ name: "", full_name: "" }];
+export type UserRepos = UserReposResponse["data"];
+export const initialRepos: UserRepos = [{ name: "", full_name: "" }];
+export type GitHubDestination = "Issue" | "File";
```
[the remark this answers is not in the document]
dvargas92495: ignore, I see the use case now. To keep in sync with the array below you could instead do:
```
const GITHUB_DESTINATIONS = ["Issue", "File"] as const;
export type GitHubDestination = (typeof GITHUB_DESTINATIONS)[number];
```

makes it so that adding new options is only write-once instead of write-twice

src/index.ts
```diff
@@ -265,6 +259,13 @@ svg.rs-svg-container {
       renderPlayground(title, globalRefs);
     } else if (isCanvasPage(title) && !!h1.closest(".roam-article")) {
       renderTldrawCanvas(title, onloadArgs);
+    } else if (isGitHubSyncPage(title)) {
```
dvargas92495, line 262: [nit] observer callbacks fire alot and roam queries within it could really slow down a graph. I would try to figure out a way to avoid doing roam queries within this method.

One way to do so is to maintain a cache of valid uids, and whenever you leave the github sync config page, that cache is updated. Then this method is only doing a lookup on the cache

src/components/GitHubSync.tsx
dvargas92495: yea this all doens't look right, this component is not mounting at all. I would replace the

```
ReactDOM.render(
      <GitHubSyncPage
        commentsContainerUid={commentsContainerUid}
        extensionAPI={extensionAPI}
      />,
      containerDiv
    );
```

from above with just the contents of this `useEffect`. Actually the "observing" is happening by the time we get to `renderGitHubSyncPage` so I wouldn't even initiate new observers for both. I would render `CommentsContainerComponent` directly there, and have a separate root observer for `CommentsComponent`

dvargas92495 (approved):
Nice! Lots of feedback, nothing necessarily blocking.

One approach I like doing with big meaty PRs is keep the meaty one in draft, then split out reviewable chunks, rebasing the master along the way. It's a ton more work, but this helps reviewers give more thoughtful reviews on incremental pieces and I've often found ways to make things better by forcing myself to chunk down.

docs/github-sync.md
```diff
@@ -0,0 +1,56 @@
+# GitHub Sync
+
+This extension implemnents GitHub Sync, allowing you to synchronize specific pages directly from Roam to a specified GitHub repository as Issues.
+
+## Config Page
+
+This will be found at `[[roam/js/github-sync]]` in your graph.
```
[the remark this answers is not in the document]
mdroidian: Initially this felt far enough away from Discourse Graph to be on it's own.  Then I ended up hooking into Discourse Nodes for their definition 😅.  But I could see users want to use only this feature and none of the other dgraph features ... then it would be kind of buried / confusing.

src/components/GitHubSync.tsx
```diff
@@ -0,0 +1,992 @@
...
+                start: 0,
+                end: 0,
+              },
+            });
+          }, 100);
+        }}
+      />
+      <Button
+        icon="import"
+        title="Check for New Comments"
```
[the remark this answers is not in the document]
mdroidian: My intuition pushes back on this.

Adding comments is likely to happen far more often than downloading.  `Check for comments` or even just `Download` places it in equal weights with `Add Comment` and it also felt like it took up too much space when I tried it.

Similarly, having them both just icons downplays `Add Comments` too much.  If both are just icons, making `Add Comments` intent `Primary` might solve this, but most of the time a user is on the page they actually won't be adding comments, so it is likely too much of a distraction.

Finally, I see a future state where we query for new comments in the background and update the button to call attention to it only when there are new comments to be imported.

src/components/GitHubSync.tsx
```diff
@@ -0,0 +1,992 @@
...
+
+  return (
+    <>
+      <Button
+        title={
+          !!url
+            ? "Comment was already sent. Click to view on GitHub."
+            : "Add to GitHub"
+        }
+        text={!!url ? "" : "Add to GitHub"}
```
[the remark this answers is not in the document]
mdroidian: I'm using the text / no text based on their importance / how the user will use them.

It's very unlikely the user will want/need to click on settings before sending the page to GitHub.  Adding that text feels like it will just distract from the main action and just take up space.

On the other hand, removing the text would leave just out of context icon(s) by their selves under the title.

src/components/GitHubSync.tsx
```diff
@@ -0,0 +1,992 @@
...
+            icon="add"
+            intent="primary"
+            onClick={async () => {
+              await Promise.all(
+                comments.map(async (c) => {
+                  const roamCreatedDate =
+                    window.roamAlphaAPI.util.dateToPageTitle(
+                      new Date(c.created_at)
+                    );
+                  const commentHeader = `${c.user.login} on [[${roamCreatedDate}]]`;
```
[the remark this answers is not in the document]
mdroidian: On the v2 list already 😁

docs/github-sync.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Node Select
+
+Select which type of pages you want to sync to GitHub as Issues.
+
+The list is made from defined [Discourse Graph Nodes](discourse-graphs.md).
+
+### Comments Block
+
+This is where the comments will live. A `Add Comment` button will appear on this block as well as a download button.
```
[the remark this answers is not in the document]
mdroidian: nice catch!

src/components/Export.tsx
mdroidian: We could store these in `localStorage` and do a check `onClick` for these values.  But if we do that for `setCanSendToGitHub` we would lose the state dependent UI of a disabled `Export` button.

src/index.ts
mdroidian: I was definitely nervous about this.  But unsure of a good solution.

I would expect that a user would visit the config page once then probably never again for a very long time.  How would the cache be updated then?  

Specifically, how would a page that is **created via default Roam methods** (`[[`, search bar, or renaming a page, for example) be added to the cache?  If it's on a timer, would the UI just not show up until that timer goes off or the user triggers a refresh?

src/components/GitHubSync.tsx
mdroidian: A few problems I'm running into trying to implement this:

**Observers Required**
Both observers are required to re-render the buttons when a user clicks into a block.

If we just render `CommentsContainerComponent` and remove the observer that renders it, if a user clicks into the block the buttons will be gone until the page reloads.  Maybe a necessary trade-off?  But I'd like to avoid that for the UX.

**RE: Root Observer**
The `CommentsComponent` renders to children of the `CommentsContainer`.  To know if it's a child we need the `CommentsContainer` `el` or `uid` which we get from a user defined query on page load.  We wouldn't have that at root.

I guess we could add an attribute on block creation/page load, but we'd run into the re-render issue still (a user clicks into the block, buttons are gone until page is reloaded).  

Or maybe we keep a running global cache of comment uids for the current page that clears on page unload ... we have to handle the sidebar windows too ... hmmm.

mdroidian: I fixed the cleanup not firing by changing `return null` to `return <></>;`

mdroidian: Here's a cache implementation: https://github.com/RoamJS/query-builder/compare/github-sync...github-sync-comment-uid-cache

mdroidian: merged into this branch: https://github.com/RoamJS/query-builder/pull/254

src/components/Export.tsx
mdroidian: https://github.com/RoamJS/query-builder/pull/255

Qiwei-Zhao:
Thank you for developing this function. It is amazing and very useful! Will we have it in the next official release?

mdroidian:
Thank you for the message @Qiwei-Zhao 

This is not yet ready for release.  Some bugs to work out and changes to be added.
Currently we have been focusing a lot on a separate upgrade, but we will be back to this shortly!

mdroidian:
Moved to https://github.com/DiscourseGraphs/discourse-graph/pull/120

PR #257, opened 2024-05-16
if share to page is new, create it

mdroidian:


src/components/Export.tsx
```diff
@@ -221,18 +223,50 @@ const ExportDialog: ExportDialogComponent = ({
...
+            } else {
+              reject(new Error("Failed to get block props"));
+            }
+          };
+
+          checkProps();
+        });
+      await waitForBlockProps();
+    }
+
```
mdroidian, line 260: not super happy with this.

a couple other methods:

- creating a `store` in a util `.ts` to populate the tldraw props with.  
  - ran into a few errors.  I believe I'd have to have the `App` context (`useApp()`) to properly create the store values I need.  I could be wrong
- just creating the `tldraw` props and putting the shapes in there
  - on page load, `pullwatch` see this as a change after it loaded the initial `tldraw` values (user, camera, etc) and overwrites those with just the added shape(s)

Let me know if you have any suggestions, or if you think it would be valuable exploring either of those options further.

dvargas92495: I'm having trouble following why we need this at all

Here's my current understanding: When you export query to a new page, you are waiting until the page loads and the tldraw store is initialized, before fetching the block props. You need the block props below to eventually just use for the `pageKey`? Is there a way to just fetch the page key?

If that's not the case and you need the full props object, it seems like you may want to calculate the props separately in a util and pass it into `createPage` (as you were mentioning in method 1). There may be a way to do so without `useApp()`

mdroidian: For method 1, Using `.createStore()` on a `new TldrawEditorConfig()` wasn't returning anything in `store.atoms.value`.  And all that was in props was `userId`, `instanceId`, and `documentId`.

So I created them manually.  https://github.com/RoamJS/query-builder/pull/257/commits/a40c93f0c516b8a0da6c3e7214694649491a947b

https://www.loom.com/share/367bdf658b4844f58c7d37ba2665e604

dvargas92495 (approved):

PR #259, opened 2024-05-18
isDiscourseNode small refactor

mdroidian:
- require `uid`, remove `nodes` params

What was the use case of having `nodes` as a param? @dvargas92495 
I checked all instances where we use `isDiscourseNode` and it looks like we are only using `uid` at the moment.

image

Related to https://github.com/RoamJS/roamjs-components/pull/26

### DIFF
--- a/docs/query-builder.md
+++ b/docs/query-builder.md
@@ -210,7 +210,7 @@ For developers of other extensions who want to use the queries defined by users,
   - `uid` - `string` The reference of the primary node involved in the result.
   - `${string}-uid` - `string` If the users define selections that return intermediary nodes, the reference of those nodes will always end in `-uid` and will always be of type `string`.
   - `{string}` - `string | number | Date` All other fields returned in the result can be any of the primitive value types.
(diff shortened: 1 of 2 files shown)


docs/query-builder.md
```diff
@@ -210,7 +210,7 @@ For developers of other extensions who want to use the queries defined by users,
   - `uid` - `string` The reference of the primary node involved in the result.
   - `${string}-uid` - `string` If the users define selections that return intermediary nodes, the reference of those nodes will always end in `-uid` and will always be of type `string`.
   - `{string}` - `string | number | Date` All other fields returned in the result can be any of the primitive value types.
-- `isDiscourseNode` - `(uid?: string, nodes?: DiscourseNode[]) => boolean` Returns whether the input `uid` is a discourse node.
+- `isDiscourseNode` - `(uid: string) => boolean` Returns whether the input `uid` is a discourse node.
```
dvargas92495, line 213: we had a previous case where we defined `nodes` outside the function and passed it down instead of calling it twice. If it's no longer true, feel free to nuke

dvargas92495 (approved):

PR #260, opened 2024-05-18
Revert "remove min width on roam settings input elements"

mdroidian:
Reverts RoamJS/query-builder#258

I rushed 258.  Not only did I merge into the wrong branch ... `.rm-settings .rm-extensions-settings input[type="text"]` would affect all extension, not just qb.  🙄


dvargas92495 (commented):
✅

PR #268, opened 2024-06-23
add support for backend querying of discourse context

mdroidian:
this is the frontend support for https://github.com/samepage-network/roam-samepage/pull/22

mostly just adds a flag, some loading state, and minor adjustments to help the backend


src/utils/getDiscourseRelations.ts
```diff
@@ -1,47 +1,50 @@
-import type { InputTextNode, TextNode } from "roamjs-components/types/native";
+import type {
```
mdroidian, line 1: `getDiscourseRelations` should just be a refactor.  It was done some time ago and note sure why I left it in this PR

PR #288, opened 2024-10-27
add command palette, add disable global

mdroidian:
`\` is often used in LaTeX so users want the ability to disable it as the global node menu trigger.

I'd probably like to phase out the global trigger in favor of the command palette, but I worry that users will request a "modifier" free option, which I don't think the command palette supports.

https://www.loom.com/share/d1ab1bb4e75446aebdb186fe39dccf5b

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,7 +52,7 @@ import { fireQuerySync } from "./utils/fireQuery";
 import parseQuery from "./utils/parseQuery";
 import { render as exportRender } from "./components/Export";
 import getPageTitleByPageUid from "roamjs-components/queries/getPageTitleByPageUid";
-
+import { render as renderDiscourseNodeMenu } from "./components/DiscourseNodeMenu";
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -391,6 +391,15 @@ svg.rs-svg-container {
...
       },
+      {
+        id: "disable-global-trigger",
+        name: "Disable Global Node Menu Trigger",
+        action: {
+          type: "switch",
+        },
+        description:
+          "Whether to disable the global trigger for the Discourse Node Menu, found at [[roam/js/discourse-graph]], and use the custom trigger instead",
+      },
```
dvargas92495, line 402: Why have a disable setting over just the custom trigger setting (where empty/obscure string effectively disables)?

mdroidian: changed to setting

dvargas92495 (approved):

PR #290, opened 2024-11-17
add browserstack to readme

mdroidian:
I dev on Windows and have been getting some css/layout issues reported by mac users that I cannot repo.

https://www.browserstack.com/open-source is free for open source if we add "This project is tested with BrowserStack".
I'd like to use their https://www.browserstack.com/live testing.


dvargas92495 (commented):
✅

PR #291, opened 2024-11-20
async.q and bypassing getDatalogQuery()

mdroidian:
proof of concept, but still TODO would be
- [ ] generate `query` / `complementQuery` on relation save
- [ ] handle `/context/` and `/anchor/` if those are still used

### DIFF
--- a/src/utils/fireQuery.ts
+++ b/src/utils/fireQuery.ts
@@ -27,6 +27,12 @@ type RelationInQuery = {
   text: string;
   isComplement: boolean;
 };
+type QuerySelection = {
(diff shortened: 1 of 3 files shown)


mdroidian:
Added a QueryTester.tsx with a bunch of results.
https://github.com/RoamJS/query-builder/pull/291/files#diff-3da15b5c2c39a89b7c2b7864e251037906a51f874ef11b71979279f9a4096c4b

The piece that is still evading me is exactly how a blocking function (`getDatalogQuery()`) inside of `fireQuery` which is wrapped in `Promise.all(queries.map(fireQuery))` changes the query response times.

